PR #20, opened 2017-03-10
Feature/ceph deploy

sven15:
* Contains update of deployment with ceph-deploy chapter (taken from develop branch)


sven15:
Why are the images updated again? They're already present in the maintenance/ses4 branch. I do not understand that.

PR #26, opened 2017-03-23
Extending chapter about cache tiering (bsc#1025189)

sven15:
* Explaining bloom filter and hit sets
* fixed hit_set_search_last_n, which was wrongly documented as hit_set_grade_search_last_n upstream


sven15:
docupdate required

PR #41, opened 2017-07-10
CIFS export of CephFS

svenseeberg:
* Samba supports vfs_ceph
* Samba failover with CTDB

### DIFF
--- a/xml/deployment_cifs.xml
+++ b/xml/deployment_cifs.xml
@@ -1,0 +1,266 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE chapter
+[
+  <!ENTITY % entities SYSTEM "entity-decl.ent">
+    %entities;
+]>
+<chapter xmlns="http://docbook.org/ns/docbook"
+ xmlns:xi="http://www.w3.org/2001/XInclude"
+ xmlns:xlink="http://www.w3.org/1999/xlink" version="5.0"
+ xml:id="cha.ses.cifs">
+<!-- ============================================================== -->
+ <title>Export &cephfs; via Samba</title>
+ <info>
+  <dm:docmanager xmlns:dm="urn:x-suse:ns:docmanager">
+   <dm:maintainer>tbazant@suse.com</dm:maintainer>
+   <dm:status>editing</dm:status>
+   <dm:deadline></dm:deadline>
+   <dm:priority></dm:priority>
+   <dm:translation></dm:translation>
+   <dm:languages></dm:languages>
+   <dm:release>SES4</dm:release>
+  </dm:docmanager>
+ </info>
+ <para>
+  This section describes how to export &cephfs; via Samba share. Samba shares
+  can be used with Windows* clients.
+ </para>
+ <warning>
+  <title>Technology Preview</title>
+  <para>
+   As of &storage; 5, <literal>CIFS Export</literal> is considered a technology
+   preview and is not supported.
+  </para>
+ </warning>
+ <sect1 xml:id="sec.ses.cifs.example">
+  <title>Example Installation</title>
+  <para>
+   Exporting &cephfs; is not a technology preview and not supported. To enable
+   CIFS, you have to manually install Samba on one cluster node and configure it.
+   Failover functionality can be provided with CTDB and the &sle; &hasi;.
+  </para>
+  <para>
+   First install the Samba daemon on one node:
+  </para>
+<screen>&prompt.root;<command>zypper</command> in samba</screen>
+  <para>
+   Then edit the <filename>/etc/smb/smb.conf</filename> and add the following
+   section:
+  </para>
+<screen>[<replaceable>SHARE_NAME</replaceable>]
+ceph:config_file = /etc/ceph/ceph.conf</screen>
+  <para>
+   Start and enable the Samba daemon:
+  </para>
+<screen>&prompt.root;<command>systemctl</command> start smb.service
+&prompt.root;<command>systemctl</command> enable smb.service</screen>
+ </sect1>
+ <sect1 xml:id="sec.ses.cifs.ha">
+  <title>High Availability Active-Passive Configuration</title>
+  <para>
+   This section provides an example of how to set up a two-ode active-passive
+   configuration of &ganesha; servers. The setup requires the &sle; &hasi;.
+   The 2 nodes are called &si.earth;
+   and &si.mars;.
+  </para>
+  <para>
+   For details about &sle; &hasi;, see
+   <link xlink:href="https://www.suse.com/documentation/sle-ha-12/" />.
+  </para>
+  <para>
+   In this setup &si.earth; has the
+   IP address <systemitem class="ipaddress">192.168.1.1</systemitem> and
+   &si.mars; has the address
+   <systemitem class="ipaddress">192.168.1.2</systemitem>.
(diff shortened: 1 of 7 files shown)


xml/deployment_cifs.xml
```diff
@@ -0,0 +1,290 @@
...
+  <para>
+   For details about &sle; &hasi;, see
+   <link xlink:href="https://www.suse.com/documentation/sle-ha-12/" />.
+  </para>
+  <para>
+   In this setup &si.earth; has the
+   IP address <systemitem class="ipaddress">192.168.1.1</systemitem> and
+   &si.mars; has the address
+   <systemitem class="ipaddress">192.168.1.2</systemitem>.
+  </para>
```
tomschr: I think this paragraph and the next paragraph could be a list like here: http://docserv.suse.de/documents/SLE-HA12-SP3/SLE-HA-install-quick/html/#sec.ha.inst.quick.usage-scenario

I would propose something like this:

--- 
The procedures in this document will lead to a minimal setup of a two-node cluster with the following properties:

* Two nodes: earth (192.168.1.1) and mars (192.168.1.2).
* Two floating, virtual IP addresses which allows clients to connect to the service no matter which physical node it is running on. <systemitem class="ipaddress">192.168.1.10</systemitem> is used for cluster administration with Hawk2 and <systemitem class="ipaddress">192.168.2.1</systemitem> is used exclusively for the CIFS exports. This makes it easier to apply security restrictions later.
* Related Samba services like CTDB, smb, and nmb.

xml/deployment_cifs.xml
```diff
@@ -0,0 +1,290 @@
...
+     Prepare the nodes used to host the Samba service:
+    </para>
+    <substeps>
+     <step>
+      <para>
+       Make sure the following packages are installed on &si.earth; and &si.mars;
+       before you proceed: <package>ctdb</package>,
+       <package>tdb-tools</package>, and <package>samba</package>
+       (needed for smb and nmb resources). 
+      </para>
```
tomschr: IMHO you don't need to repeat the packages inside the paragraph. It's enough to have that inside the screen. I would suggest to rephrase it like:

```xml
<para>Make sure the following packages are installed on &si.earth; and &si.mars;:</para>
<screen>...</screen>
```

xml/deployment_cifs.xml
```diff
@@ -0,0 +1,290 @@
...
+<screen>&prompt.root;<command>zypper</command> in ctdb tdb-tools samba</screen>
+     </step>
+     <step>
+      <para>
+       Make sure the services <literal>ctdb</literal>, <literal>smb</literal>,
+       and <literal>nmb</literal> are disabled:
```
tomschr: If you like, you can put the reason behind: "... as they are managed by Pacemaker:"

xml/deployment_cifs.xml
```diff
@@ -0,0 +1,290 @@
...
+  </para>
+  <para>
+   The following procedure describes the example installation. More details
+   can be found at <link xlink:href=
+   "https://www.suse.com/documentation/sle-ha-12/install-quick/data/install-quick.html" />.
+  </para>
+  <procedure xml:id="proc.sec.ses.cifs.ha">
+   <step>
+    <para>
+     Prepare the nodes used to host the Samba service:
```
tomschr: It's not clear on which nodes you need to execute this. So maybe rephrase it like "On both nodes, prepare the Samba service(s):"

xml/deployment_cifs.xml
```diff
@@ -0,0 +1,290 @@
...
+       Open port <literal>4379</literal> of your firewall on all nodes. This
+       is needed for CTDB to communicate with other cluster nodes.
+      </para>
+     </step>
+    </substeps>
+   </step>
+   <step>
+    <para>
+     Create a directory for the CTDB lock on the shared file system:
+    </para>
```
tomschr: Maybe move this step as a substep to step 1.

xml/deployment_cifs.xml
```diff
@@ -0,0 +1,290 @@
...
+<screen>&prompt.root;<command>mkdir</command> -p /srv/samba/</screen>
+   </step>
+   <step>
+    <para>
+     In <filename>/etc/ctdb/nodes</filename> insert all nodes which contain
+     all private IP addresses of each node in the cluster:
+    </para>
+<screen>&subnetI;.1
+&subnetI;.2</screen>
```
tomschr: Maybe move this step as a substep to step 1.

xml/deployment_cifs.xml
```diff
@@ -0,0 +1,290 @@
...
+   <step>
+    <para>
+     Install <package>ha-cluster-bootstrap</package> on both nodes:
+    </para>
+    <screen>
+&prompt.root;<command>zypper</command> in ha-cluster-bootstrap
+    </screen>
+   </step>
+   <step>
+    <substeps>
```
tomschr: Maybe combine this substep with the last step to create something like this:

---

3. Set up your two-node cluster:
  a. Install ha-cluster-bootstrap on both nodes: 
  b. Create a one-node cluster on earth:
  c. Let mars join the cluster:

xml/deployment_cifs.xml
```diff
@@ -0,0 +1,290 @@
...
+     <step>
+      <para>
+       Init the cluster on &si.earth;:
```
tomschr: "Init" is jargon. 😉 Better rephrase it like "Create a one-node cluster on &si.earth;:"

xml/deployment_cifs.xml
```diff
@@ -0,0 +1,290 @@
...
+     </step>
+    </substeps>
+   </step>
+   <step>
+    <para>
+     Check the status of cluster. You should see two nodes added in cluster:
+    </para>
+    <screen>
+&prompt.earth;<command>crm</command> status
+    </screen>
```
tomschr: If you like, you could add the output of `crm status`:

```
2 nodes configured
1 resource configured

Online: [ earth mars ]

Full list of resources:

 admin-ip       (ocf::heartbeat:IPaddr2):       Started earth
```

xml/deployment_cifs.xml
```diff
@@ -0,0 +1,290 @@
...
+   <step>
+    <para>
+     Execute the following commands on &si.earth; to configure the the CTDB resource:
+    </para>
+<!--taroth 2012-02-07: fixed bnc#745334, default timeout for start/stop-->
+<screen>&prompt.earth;<command>crm</command> configure
+&prompt.crm.conf;<command>primitive</command> ctdb ocf:heartbeat:CTDB params \
+    ctdb_manages_winbind="false" \ 
+    ctdb_manages_samba="false" \
+    ctdb_recovery_lock="/srv/samba/ctdb.lock" \
```
tomschr: Maybe something for developers to answer: I'm not sure, but should the lock somewhere in `/var/lock/....`?

xml/deployment_cifs.xml
```diff
@@ -60,7 +60,8 @@
    </step>
    <step>
     <para>
-     Install the Samba daemon on the Samba gateway node:
+     The follwing steps are executed on the Samba gateway node. Install the
```
ddiss: Typo: following

xml/deployment_cifs.xml
svenseeberg: Done

PR #50, opened 2017-12-14
Move tcmu-runner service status check to installation section

ricardoasmarques:
If the tcmu-runner service is not running, the first step described on "**9.5.2 Configuration and Deployment**" will fail because tcmu-runner RBD handler requires this service to be running.

For this reason, this PR moves the service status check from the bottom of the section "**9.5.2 Configuration and Deployment**" to the bottom of "**9.5.1 Installation**" section.

Signed-off-by: Ricardo Marques <rimarques@suse.com>


svenseeberg:
Thanks for the PR!

PR #51, opened 2017-12-18
Clarify that admin node should be upgraded first

smithfarm:
Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1073277
Signed-off-by: Nathan Cutler <ncutler@suse.com>

### DIFF
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -624,6 +624,11 @@ exported keyring for client.bootstrap-osd
      </para>
     </listitem>
    </itemizedlist>
+   <para>
+    In addition, before starting the procedure below, you need to upgrade the
+    admin node to &sls; 12 SP3 and &storage; 5 by running <command>zypper
+    migration</command> (or your preferred way of upgrading).
+   </para>
   </important>
 
   <para>
@@ -635,22 +640,14 @@ exported keyring for client.bootstrap-osd
    <title>Steps to Apply to All Cluster Nodes</title>
    <step>
     <para>
-     If you use SCC-registered systems, no actions have to be taken. Otherwise
-     prepare software repositories. Disable all repositories except for &sls;
-     12 SP3 and &storage; 5 repositories. If these repositories are missing,
-     add them.
-    </para>
-   </step>
-   <step>
-    <para>
-     Install the <systemitem>salt</systemitem> package:
+     Install the <systemitem>salt</systemitem> package from SLE-12-SP2/SES4:
     </para>
 <screen>&prompt.root;zypper install salt</screen>
    </step>
    <step>
     <para>
-     Install the <systemitem>salt-minion</systemitem> package, then enable and
-     start the related service:
+     Install the <systemitem>salt-minion</systemitem> package from
+     SLE-12-SP2/SES4, then enable and start the related service:
     </para>
 <screen>&prompt.root;zypper install salt-minion
 &prompt.root;systemctl enable salt-minion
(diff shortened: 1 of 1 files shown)


smithfarm:
@tbazant @svenseeberg @Martin-Weiss

Martin-Weiss:
FYI - had a look again and it seems that the whole process is having a wrong order for:

<SNIP>
Steps to Apply to All Cluster Nodes

1.    If you use SCC-registered systems, no actions have to be taken. Otherwise prepare software repositories. Disable all repositories except for SUSE Linux Enterprise Server 12 SP3 and SUSE Enterprise Storage 5 repositories. If these repositories are missing, add them. 
</SNIP>

If I do that for all servers in the cluster  they will all install salt packages from SLES 12 SP3 channels onto SLES 12 SP2 servers ! I assume that this is NOT supported at all.

I believe the correct process would be to install the salt-minions from the SLES 12 SP2 channels.. and attach the SLES 12 SP3 / SES 5 channels after that.

smithfarm:
@Martin-Weiss Good point! I pushed a new version of the fix - please take another look.

xml/admin_ceph_upgrade.xml
```diff
@@ -624,6 +624,11 @@ exported keyring for client.bootstrap-osd
      </para>
     </listitem>
    </itemizedlist>
+   <para>
+    In addition, before starting the procedure below, you need to upgrade the
+    &sadmin; node to &sls; 12 SP3 and &storage; 5 by running <command>zypper
```
svenseeberg: We have no sadmin entity. You can either add it to the entity-decl.ent or replace \&sadmin; with \&salt; master.

xml/admin_ceph_upgrade.xml
```diff
@@ -689,6 +686,14 @@ exported keyring for client.bootstrap-osd
  "flags": "sortbitwise,recovery_deletes,purged_snapdirs",</screen>
     </tip>
    </step>
+   <step>
+    <para>
+     Upgrade the &sadmin; node to &sls; 12 SP3 and &storage; 5. If you use
+     SCC-registered systems, we recommend <command>zypper migration</command>.
+     After the upgrade, ensure that only repositories for &sls; 12 SP3 and
+     &storage; 5 are active (and refreshed) on the &sadmin; node before proceeding.
```
svenseeberg: We have no sadmin entity. You can either add it to the entity-decl.ent or replace \&sadmin; with \&salt; master.

xml/admin_ceph_upgrade.xml
```diff
@@ -689,6 +686,14 @@ exported keyring for client.bootstrap-osd
  "flags": "sortbitwise,recovery_deletes,purged_snapdirs",</screen>
     </tip>
    </step>
+   <step>
+    <para>
+     Upgrade the &sadmin; node to &sls; 12 SP3 and &storage; 5. If you use
```
svenseeberg: We have no sadmin entity. You can either add it to the entity-decl.ent or replace \&sadmin; with \&salt; master.

smithfarm:
@svenseeberg Done

@Martin-Weiss Please take another look.

smithfarm:
@jschmid1 How does this look to you?

xml/admin_ceph_upgrade.xml
```diff
@@ -624,6 +624,11 @@ exported keyring for client.bootstrap-osd
      </para>
     </listitem>
    </itemizedlist>
+   <para>
```
jschmid1: I think that'd tell the user to upgrade the admin _before_ setting up a salt cluster.
AFAIK the procedure starts with a SP2 based salt cluster, then upgrade the admin to SP3 in order to get the latest code which includes the engulfment

smithfarm: If you have a SP2-based salt cluster, and then upgrade the master node to SP3, that becomes a mixed Salt cluster (salt-master and salt-minion on the master node will be from SP3/SES5, and the rest of the minions will be SP2/SES4). That is the starting point for the engulfment and the maintenance.upgrade orchestration that follows.

svenseeberg:
Feel free to merge this PR. I have no objections.

xml/admin_ceph_upgrade.xml
jschmid1: ok, fine with me.

PR #52, opened 2018-01-18
remove osd comment and saftey.disangage

markostanojlovic:
adding safty disangage command line that is required 
giving the description about how to specify osd_ID


svenseeberg:
I made some changes and created new PR #54

PR #68, opened 2019-04-17
Typo when accepting salt-key

SweBarre:
When accepting keys with salt-key it is either `salt-key -a` or `salt-key --accept`
The document said `salt-key --a` and that throws a `salt-key: error: ambiguous option: --a`

I do think the long options are better suitable for documentation.


tbazant:
Thanks, you are right

PR #91, opened 2019-06-18
jsc#SES-178 clarify when Samba oplocks / leases are safe to use

ddiss:


xml/deployment_cifs.xml
```diff
@@ -131,9 +131,13 @@
       <para>
        <option>oplocks</option> (also known as SMB2+ leases) allow for improved
        performance through aggressive client caching, but are currently unsafe
-       when Samba is deployed together with other &cephfs; clients, such as
+       when &samba; is deployed together with other &cephfs; clients, such as
        kernel <literal>mount.cephfs</literal>, FUSE, or &ganesha;.
       </para>
+      <para>
+       If all &cephfs; file system path access is exclusively handled by
```
tbazant:
```suggestion
       If all &cephfs; file system paths' access is exclusively handled by
```

ddiss: Hmm, I find this a little confusing. Basically the point I'm trying to convey is:

Any CephFS path exposed as a Samba share should only be accessed via Samba (from an SMB client). This includes all subdirectories and files under the Samba share path. If that is the case, then the <option>oplocks</option> parameter can be safely enabled for the share.

tbazant: so....maybe leaving it your way is a solution

PR #105, opened 2019-10-21
Add "run discovery stage command" instruction (BSC#1154437)

asettle:
The initial step 5 did not clarify that you are required to run
the following commands to initiate the discovery phase. This
addition syncs the instructions with step 7, and makes it clear
to the user they have to run these commands.


asettle:
Requires backport to maintenance/ses6

asettle:
Eric approved: https://bugzilla.suse.com/show_bug.cgi?id=1154437#c4

PR #108, opened 2019-10-22
Introduced an automatic way of migrating a node during cluster upgrade (bsc#1154438)

tbazant:
- added the mentioned section
- updated the docupdates section


xml/admin_ceph_upgrade.xml
```diff
@@ -931,83 +933,69 @@ mds standby for name = mds.456-another-mds
...
+    <step>
+     <para>
+      Reboot to start the upgrade. The node will reboot automatically after the
+      upgrade.
+     </para>
+     <tip>
+      <title>Upgrade Failure</title>
+      <para>
+       If the upgrade fails, inspect
+       <filename>/var/log/distro_migration.log</filename>. Then fix the
```
asettle:
```suggestion
       <filename>/var/log/distro_migration.log</filename>. Fix the

File: xml/admin_ceph_upgrade.xml [hunk 2]
```diff
@@ -1595,9 +1583,11 @@ role-storage/cluster/*.sls
 <screen>&prompt.smaster;salt-run state.orch ceph.restart.rgw.force</screen>
    </step>
    <step>
-     <para>Before you continue, we strongly recommend enabling the Ceph
-       telemetry module. For more information, see <xref linkend="mgr-modules-telemetry"/>
-       for information and instructions.</para>
+    <para>
+     Before you continue, we strongly recommend enabling the Ceph telemetry
```
asettle:
```suggestion
     Before you continue, we strongly recommend enabling the &ceph; telemetry

**Comment by tbazant** [comment]
**Do not merge this PR before  https://maintenance.suse.de/incident/12912 and https://maintenance.suse.de/incident/12672 incidents are resolved**

File: xml/admin_ceph_upgrade.xml [hunk 3]
```diff
@@ -931,83 +951,93 @@ mds standby for name = mds.456-another-mds
...
+        &susemgr;, create the
+        <filename>/etc/sle-migration-service.yml</filename> with the following
+        content:
+       </para>
 <screen>
-&prompt.root;zypper mr -d -a
+use_zypper_migration: true
+preserve:
+rules:
+- /etc/udev/rules.d/70-persistent-net.rules
```
tserong: `rules:` and the subsequent line need one more indent as under "Preserve System Data" at https://documentation.suse.com/suse-distribution-migration-system/1.0/single-html/distribution-migration-system/#_optional_customization_of_the_upgrade_process, so this should be:

```
use_zypper_migration: true
preserve:
  rules:
    - /etc/udev/rules.d/70-persistent-net.rules
```

(Also, `use_zypper_migration: true` is the default setting, to technically doesn't need to be specified in this case, but there's no harm in leaving it there for symmetry with the other case)

xml/admin_ceph_upgrade.xml
```diff
@@ -931,83 +951,93 @@ mds standby for name = mds.456-another-mds
...
+         <para>
+          Create the <filename>/etc/sle-migration-service.yml</filename> with
+          the following content:
+         </para>
 <screen>
-zypper ar -f <replaceable>REPO_URL</replaceable>
+use_zypper_migration: false
+preserve:
+rules:
+- /etc/udev/rules.d/70-persistent-net.rules
```
tserong: Same comment as a above: re indentation

xml/admin_ceph_upgrade.xml
tbazant: updated

xml/admin_ceph_upgrade.xml
tbazant: updated

xml/admin_ceph_upgrade.xml
```diff
@@ -931,83 +951,93 @@ mds standby for name = mds.456-another-mds
...
+         </para>
+        </step>
+       </substeps>
+      </step>
+     </substeps>
+    </step>
+    <step>
+     <para>
+      Reboot to start the upgrade. The node will reboot automatically after the
+      upgrade.
```
tserong: Please add something like "While the upgrade is running, it's possible to ssh in as the migration user, using an existing ssh key from the host system, as described in the DMS documentation. For SES, it is also possible to log in as root on the system console, with the password "sesupgrade" (i.e. if you have physical access to the machine, or direct console access)".

tserong:
> Do not merge this PR before https://maintenance.suse.de/incident/12912 and https://maintenance.suse.de/incident/12672 incidents are resolved

Those MRs are done now, so we're good to merge (pending my other comment above about console access)

PR #111, opened 2019-10-22
Added min_size for EC pools (BSC#1153481)

asettle:
The default for min_size on EC pools is k+1.
As changed here: https://github.com/ceph/ceph/pull/8008

### DIFF
--- a/xml/admin_ceph_erasure.xml
+++ b/xml/admin_ceph_erasure.xml
@@ -154,9 +154,10 @@ ABCDEFGHI</screen>
     <listitem>
      <para>
       the number of data chunks, that is the number of chunks into which the
-      original object is divided. For example if <literal>k = 2</literal> a
+      original object is divided. For example, if <literal>k = 2</literal> a
       10KB object will be divided into <literal>k</literal> objects of 5KB
       each.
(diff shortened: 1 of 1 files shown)


xml/admin_ceph_erasure.xml
```diff
@@ -154,9 +154,10 @@ ABCDEFGHI</screen>
     <listitem>
      <para>
       the number of data chunks, that is the number of chunks into which the
-      original object is divided. For example if <literal>k = 2</literal> a
+      original object is divided. For example, if <literal>k = 2</literal> a
       10KB object will be divided into <literal>k</literal> objects of 5KB
       each.
+      The default <literal>min_size</literal> on erasure pools is <literal>k + 1</literal>.
```
tbazant:
```suggestion
      The default <literal>min_size</literal> on erasure coded pools is <literal>k + 1</literal>.
```

asettle:
> LGTM
> 
> Though if one wants to be precise, for a k+m profile the default min_size is k+1 only if m > 1. If m is 1, the default min_size is k so we would allow it function when one shard is down. However I doubt someone uses m=1 in production so I am not sure it worth mentioning.

It's good to know though. I do think it might be too much info, but I see you've added this to the BZ in case it comes up again :)

PR #113, opened 2019-10-22
Removing "bytes" confusion from '5G' example (BSC#1153582)

asettle:
The original example confused the reader by mentioning "bytes".
5G is not bytes, and the recommendation follows the example
upstream. See: https://github.com/SUSE/DeepSea/wiki/Drive-Groups#specification

### DIFF
--- a/xml/admin_install_salt.xml
+++ b/xml/admin_install_salt.xml
@@ -1493,8 +1493,8 @@ drive_group_default_name:
     drive_spec: <replaceable>DEVICE_SPECIFICATION</replaceable>
   wal_devices:
     drive_spec: <replaceable>DEVICE_SPECIFICATION</replaceable>
-  block_wal_size: '5G'  # (optional)
-  block_db_size: '5G'   # (optional)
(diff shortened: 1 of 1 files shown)


xml/admin_install_salt.xml
```diff
@@ -1493,8 +1493,8 @@ drive_group_default_name:
     drive_spec: <replaceable>DEVICE_SPECIFICATION</replaceable>
   wal_devices:
     drive_spec: <replaceable>DEVICE_SPECIFICATION</replaceable>
-  block_wal_size: '5G'  # in bytes, you can use shortcuts; optional
-  block_db_size: '5G'   # in bytes, you can use shortcuts; optional
+  block_wal_size: '5G'  # (optional)
+  block_db_size: '5G'   # (optional)
```
tbazant: consider adding something like 'unit suffixes permitted', but if you think it's clear from the example - just ignore this comment

asettle:
> consider adding something like 'unit suffixes permitted', but if you think it's clear from the example - just ignore this comment

I think that's a good suggestion. Adding!

PR #117, opened 2019-10-23
Adding filestore to bluestore upgrade content (BSC#1141175)

asettle:
Bluestore is the default for SES 5.x and 6, however for customers
performing any upgrade path, they may not have completed this
requirement when upgrading from SES 4 to SES 5. This PR
includes the content from SES 5.5 docs.

### DIFF
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -1402,7 +1402,7 @@ data:
   <title>OSD Migration to &bluestore;</title>
   <para>
    OSD &bluestore; is a new back end for the OSD daemons. It is the default
-   option since &productname; &productnumber;. Compared to FileStore, which
+   option since &productname; 5. Compared to &filestore;, which
    stores objects as files in an XFS file system, &bluestore; can deliver
    increased performance because it stores objects directly on the underlying
    block device. &bluestore; also enables other features, such as built-in
@@ -1412,20 +1412,20 @@ data:
    Specifically for &bluestore;, an OSD has a 'wal' (Write Ahead Log) device
    and a 'db' (RocksDB database) device. The RocksDB database holds the
    metadata for a &bluestore; OSD. These two devices will reside on the same
-   device as an OSD by default, but either can be placed on faster/different
-   media.
+   device as an OSD by default, but either can be placed on different, for
+   example faster, media.
   </para>
   <para>
-   In SES5, both FileStore and &bluestore; are supported and it is possible
-   for FileStore and &bluestore; OSDs to co-exist in a single cluster. During
-   the SUSE Enterprise Storage upgrade procedure, FileStore OSDs are not
+   In &productname; 5, both &filestore; and &bluestore; are supported and it is possible
+   for &filestore; and &bluestore; OSDs to co-exist in a single cluster. During
+   the &productname; upgrade procedure, &filestore; OSDs are not
    automatically converted to &bluestore;. Be aware that the
    &bluestore;-specific features will not be available on OSDs that have not
    been migrated to &bluestore;.
   </para>
   <para>
-   Before converting to &bluestore;, the OSDs need to be running &productname;
-   &productnumber;. The conversion is a slow process as all data gets
+   Before converting to &bluestore;, the OSDs need to be running &productname; 5.
+   The conversion is a slow process as all data gets
    re-written twice. Though the migration process can take a long time to
    complete, there is no cluster outage and all clients can continue accessing
    the cluster during this period. However, do expect lower performance for
@@ -1433,7 +1433,7 @@ data:
    backfilling of cluster data.
   </para>
   <para>
-   Use the following procedure to migrate FileStore OSDs to &bluestore;:
+   Use the following procedure to migrate &filestore; OSDs to &bluestore;:
   </para>
   <tip>
    <title>Turn Off Safety Measures</title>
@@ -1451,17 +1451,18 @@ data:
  </screen>
    <para>You can also choose to rebuild each node individually. For example:</para>
  <screen>
-   &prompt.smaster; salt-run rebuild.node data1.ceph
+&prompt.smaster; salt-run rebuild.node data1.ceph
  </screen>
-   <para>The rebuild is not idempotent. To address the complication of shared
-     devices, the first step of a rebuild is to remove all the OSDs on a node.
-     If one OSD fails to convert, rerunning the rebuild will destroy the
-     already converted Bluestore OSDs. Depending on the failure and the
-     amount of progress, for example we recommend running:</para>
- <screen>
-   &prompt.smaster; salt-run disks.deploy target=data1.ceph
- </screen>
-  </tip>
+   <para>The <literal>rebuild.node</literal> always removes and recreates
+     all OSDs on the node.</para>
+     <important>
+       <para>If one OSD fails to convert, re-running the rebuild destroys the
+         already converted &bluestore; OSDs. If this happens, run:</para>
+         <screen>
+           &prompt.smaster; salt-run disks.deploy <replaceable>TARGET</replaceable>
+         </screen>
+       </important>
+ </tip>
   <procedure>
    <step>
     <para>
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -1406,7 +1406,7 @@ data:
    stores objects as files in an XFS file system, &bluestore; can deliver
    increased performance because it stores objects directly on the underlying
    block device. &bluestore; also enables other features, such as built-in
-   compression and EC overwrites, that are unavailable with FileStore.
+   compression and EC overwrites, that are unavailable with &filestore.
   </para>
   <para>
    Specifically for &bluestore;, an OSD has a 'wal' (Write Ahead Log) device
@@ -1475,7 +1475,7 @@ data:
      <filename>policy.cfg</filename>, finds any hardware profile using the
      original data structure, and converts it to the new data structure. The
      result is a new hardware profile named
-     'migrated-<replaceable>original_name</replaceable>'.
+     'migrated-<replaceable>ORIGINAL_NAME</replaceable>'.
      <filename>policy.cfg</filename> is updated as well.
     </para>
     <para>
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -1397,6 +1397,146 @@ data:
    </step>
   </procedure>
  </sect1>
+
+ <sect1 xml:id="filestore2bluestore">
+  <title>OSD Migration to &bluestore;</title>
+  <para>
+   OSD &bluestore; is a new back end for the OSD daemons. It is the default
+   option since &productname; &productnumber;. Compared to FileStore, which
+   stores objects as files in an XFS file system, &bluestore; can deliver
+   increased performance because it stores objects directly on the underlying
+   block device. &bluestore; also enables other features, such as built-in
+   compression and EC overwrites, that are unavailable with FileStore.
+  </para>
+  <para>
+   Specifically for &bluestore;, an OSD has a 'wal' (Write Ahead Log) device
+   and a 'db' (RocksDB database) device. The RocksDB database holds the
+   metadata for a &bluestore; OSD. These two devices will reside on the same
+   device as an OSD by default, but either can be placed on faster/different
+   media.
+  </para>
+  <para>
+   In SES5, both FileStore and &bluestore; are supported and it is possible
+   for FileStore and &bluestore; OSDs to co-exist in a single cluster. During
+   the SUSE Enterprise Storage upgrade procedure, FileStore OSDs are not
+   automatically converted to &bluestore;. Be aware that the
+   &bluestore;-specific features will not be available on OSDs that have not
+   been migrated to &bluestore;.
+  </para>
+  <para>
+   Before converting to &bluestore;, the OSDs need to be running &productname;
+   &productnumber;. The conversion is a slow process as all data gets
+   re-written twice. Though the migration process can take a long time to
+   complete, there is no cluster outage and all clients can continue accessing
+   the cluster during this period. However, do expect lower performance for
+   the duration of the migration. This is caused by rebalancing and
+   backfilling of cluster data.
+  </para>
+  <para>
+   Use the following procedure to migrate FileStore OSDs to &bluestore;:
+  </para>
+  <tip>
+   <title>Turn Off Safety Measures</title>
+   <para>
+    &salt; commands needed for running the migration are blocked by safety
+    measures. In order to turn these precautions off, run the following
+    command:
+   </para>
+ <screen>
+ &prompt.smaster;salt-run disengage.safety
+ </screen>
+   <para>Rebuild the nodes before continuing:</para>
+ <screen>
+ &prompt.smaster; salt-run rebuild.node <replaceable>TARGET</replaceable>
+ </screen>
+   <para>You can also choose to rebuild each node individually. For example:</para>
+ <screen>
+   &prompt.smaster; salt-run rebuild.node data1.ceph
+ </screen>
+   <para>The rebuild is not idempotent. To address the complication of shared
+     devices, the first step of a rebuild is to remove all the OSDs on a node.
+     If one OSD fails to convert, rerunning the rebuild will destroy the
+     already converted Bluestore OSDs. Depending on the failure and the
+     amount of progress, for example we recommend running:</para>
+ <screen>
+   &prompt.smaster; salt-run disks.deploy target=data1.ceph
+ </screen>
+  </tip>
+  <procedure>
+   <step>
+    <para>
+     Migrate hardware profiles:
+    </para>
+ <screen>&prompt.smaster;salt-run state.orch ceph.migrate.policy</screen>
+    <para>
+     This runner migrates any hardware profiles currently in use by the
+     <filename>policy.cfg</filename> file. It processes
+     <filename>policy.cfg</filename>, finds any hardware profile using the
+     original data structure, and converts it to the new data structure. The
+     result is a new hardware profile named
+     'migrated-<replaceable>original_name</replaceable>'.
+     <filename>policy.cfg</filename> is updated as well.
+    </para>
+    <para>
+     If the original configuration had separate journals, the &bluestore;
+     configuration will use the same device for the 'wal' and 'db' for that
+     OSD.
+    </para>
+   </step>
+   <step>
+    <para>
+     &deepsea; migrates OSDs by setting their weight to 0 which 'vacuums' the
+     data until the OSD is empty. You can either migrate OSDs one by one, or
+     all OSDs at once. In either case, when the OSD is empty, the
+     orchestration removes it and then re-creates it with the new
+     configuration.
+    </para>
+    <tip>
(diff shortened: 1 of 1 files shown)


swiftgist:
With SES5, we are using Salt orchestrations.  In SES6, we switched to using a runner.  

The main weakness of migrating one OSD at a time is when the administrator decides to change the sizing significantly and switch from Filestore to Bluestore.  Effectively, the migration is not intelligent and does not resolve issues for shared devices.  The workaround has been to remove all the OSDs from a node and redeploy the node.  

The first four paragraphs can be used for SES6.  The command is

`# salt-run disengage.safety`
and
`# salt-run rebuild.node TARGET`

where TARGET can be any valid salt target.  For converting from Filestore to Bluestore, run

`# salt-run rebuild.node I@roles:storage`

For those wishing to take a more conservative transition, each node could be rebuilt individually.  For instance, 

`# salt-run rebuild.node data1.ceph`

---
One important difference between the migration in SES5 and the rebuld in SES6.  The migration is idempotent.  In other words, when an OSD fails to convert for any reason, restarting the migration will pickup where it left off.

The rebuild is **not** idempotent.  To address the complication of shared devices, the first step of a rebuild is to remove all the OSDs on a node.  If one OSD fails to convert, rerunning the rebuild will destroy the already converted Bluestore OSDs.  Depending on the failure and the amount of progress, the normal resolution is to run 

`# salt-run disks.deploy target=data1.ceph`

to deploy any remaining OSDs.

asettle:
> With SES5, we are using Salt orchestrations. In SES6, we switched to using a runner.
> 
> The main weakness of migrating one OSD at a time is when the administrator decides to change the sizing significantly and switch from Filestore to Bluestore. Effectively, the migration is not intelligent and does not resolve issues for shared devices. The workaround has been to remove all the OSDs from a node and redeploy the node.
> 
> The first four paragraphs can be used for SES6. The command is
> 
> `# salt-run disengage.safety`
> and
> `# salt-run rebuild.node TARGET`
> 
> where TARGET can be any valid salt target. For converting from Filestore to Bluestore, run
> 
> `# salt-run rebuild.node I@roles:storage`
> 
> For those wishing to take a more conservative transition, each node could be rebuilt individually. For instance,
> 
> `# salt-run rebuild.node data1.ceph`
> 
> One important difference between the migration in SES5 and the rebuld in SES6. The migration is idempotent. In other words, when an OSD fails to convert for any reason, restarting the migration will pickup where it left off.
> 
> The rebuild is **not** idempotent. To address the complication of shared devices, the first step of a rebuild is to remove all the OSDs on a node. If one OSD fails to convert, rerunning the rebuild will destroy the already converted Bluestore OSDs. Depending on the failure and the amount of progress, the normal resolution is to run
> 
> `# salt-run disks.deploy target=data1.ceph`
> 
> to deploy any remaining OSDs.

Hey @swiftgist ,

Thanks for this! So if I'm understanding correctly, after running `salt-run disengage.safety` you *also* have to run `salt-run rebuild.node` to rebuild the node before you migrate? Is that correct?

swiftgist:
> 
> Hey @swiftgist ,
> 
> Thanks for this! So if I'm understanding correctly, after running `salt-run disengage.safety` you _also_ have to run `salt-run rebuild.node` to rebuild the node before you migrate? Is that correct?

Right.  

A quick explanation about `disengage.safety`.  Salt cannot generally be interactive since jobs are executed asynchronously.  Rather than try to force the "Are you sure? [Y/N]" throughout, I opted for running a prerequisite command.  Other commands such as `rebuild.node` check if that command has been run in the previous minute.

The main purpose for this is to prevent accidental reruns of the same command.  Imagine a tired admin hitting the up arrow too many times, pressing enter at the wee hours of the morning and accidentally rebuilding a working storage node.  Hitting Ctrl-C will not stop any jobs already sent to Salt.  The end result could be waiting for the rebuild to complete, which may take hours or killing the Salt jobs and then recovering from there.  The current behavior enforces the other option which is to make this a two step process as a form of confirmation.

asettle:
Thanks @swiftgist :+1: I have updated, let me know if I have misunderstood anything.

xml/admin_ceph_upgrade.xml
```diff
@@ -1397,6 +1397,146 @@ data:
    </step>
   </procedure>
  </sect1>
+
+ <sect1 xml:id="filestore2bluestore">
+  <title>OSD Migration to &bluestore;</title>
+  <para>
+   OSD &bluestore; is a new back end for the OSD daemons. It is the default
+   option since &productname; &productnumber;. Compared to FileStore, which
```
lproven: I suspect it does not make sense to specify a particular capability since a specific version of a product (any product) using an updating entity. Surely this is a fixed, specific release and should be explicitly identified?

xml/admin_ceph_upgrade.xml
```diff
@@ -1397,6 +1397,146 @@ data:
...
+   increased performance because it stores objects directly on the underlying
+   block device. &bluestore; also enables other features, such as built-in
+   compression and EC overwrites, that are unavailable with FileStore.
+  </para>
+  <para>
+   Specifically for &bluestore;, an OSD has a 'wal' (Write Ahead Log) device
+   and a 'db' (RocksDB database) device. The RocksDB database holds the
+   metadata for a &bluestore; OSD. These two devices will reside on the same
+   device as an OSD by default, but either can be placed on faster/different
+   media.
```
lproven: Suggest: "on different, for example faster, media"

xml/admin_ceph_upgrade.xml
```diff
@@ -1397,6 +1397,146 @@ data:
...
+   In SES5, both FileStore and &bluestore; are supported and it is possible
+   for FileStore and &bluestore; OSDs to co-exist in a single cluster. During
+   the SUSE Enterprise Storage upgrade procedure, FileStore OSDs are not
+   automatically converted to &bluestore;. Be aware that the
+   &bluestore;-specific features will not be available on OSDs that have not
+   been migrated to &bluestore;.
+  </para>
+  <para>
+   Before converting to &bluestore;, the OSDs need to be running &productname;
+   &productnumber;. The conversion is a slow process as all data gets
```
lproven: Does this need to be an explicit, specific version, too?

xml/admin_ceph_upgrade.xml
```diff
@@ -1397,6 +1397,146 @@ data:
...
+  <tip>
+   <title>Turn Off Safety Measures</title>
+   <para>
+    &salt; commands needed for running the migration are blocked by safety
+    measures. In order to turn these precautions off, run the following
+    command:
+   </para>
+ <screen>
+ &prompt.smaster;salt-run disengage.safety
+ </screen>
```
lproven: I have been **sternly** told off for newlines after <screen> or before </screen>

xml/admin_ceph_upgrade.xml
```diff
@@ -1397,6 +1397,146 @@ data:
...
+   <para>Rebuild the nodes before continuing:</para>
+ <screen>
+ &prompt.smaster; salt-run rebuild.node <replaceable>TARGET</replaceable>
+ </screen>
+   <para>You can also choose to rebuild each node individually. For example:</para>
+ <screen>
+   &prompt.smaster; salt-run rebuild.node data1.ceph
+ </screen>
+   <para>The rebuild is not idempotent. To address the complication of shared
```
lproven: Two things.

1. I think the word "idempotent" needs to be explained.
2. Doesn't this contradict @swiftgist's statement in the comments?

xml/admin_ceph_upgrade.xml
```diff
@@ -1397,6 +1397,146 @@ data:
...
+     devices, the first step of a rebuild is to remove all the OSDs on a node.
+     If one OSD fails to convert, rerunning the rebuild will destroy the
+     already converted Bluestore OSDs. Depending on the failure and the
+     amount of progress, for example we recommend running:</para>
```
lproven: "Depending on the $thing and the $otherthing, we recommend..."

Depending on what? Does it depend on it having not got very far, or on having it nearly finished? If it depends on a type of failure, *what* type of failure?

xml/admin_ceph_upgrade.xml
asettle: This is the way @tbazant wanted the screen tags to be indented.

xml/admin_ceph_upgrade.xml
asettle: @swiftgist ?

xml/admin_ceph_upgrade.xml
```diff
@@ -1397,6 +1397,146 @@ data:
...
+   Specifically for &bluestore;, an OSD has a 'wal' (Write Ahead Log) device
+   and a 'db' (RocksDB database) device. The RocksDB database holds the
+   metadata for a &bluestore; OSD. These two devices will reside on the same
+   device as an OSD by default, but either can be placed on faster/different
+   media.
+  </para>
+  <para>
+   In SES5, both FileStore and &bluestore; are supported and it is possible
+   for FileStore and &bluestore; OSDs to co-exist in a single cluster. During
+   the SUSE Enterprise Storage upgrade procedure, FileStore OSDs are not
```
lproven: Surely there's an entity?

xml/admin_ceph_upgrade.xml
```diff
@@ -1397,6 +1397,146 @@ data:
...
+  </para>
+  <para>
+   Specifically for &bluestore;, an OSD has a 'wal' (Write Ahead Log) device
+   and a 'db' (RocksDB database) device. The RocksDB database holds the
+   metadata for a &bluestore; OSD. These two devices will reside on the same
+   device as an OSD by default, but either can be placed on faster/different
+   media.
+  </para>
+  <para>
+   In SES5, both FileStore and &bluestore; are supported and it is possible
```
lproven: Entity here?

asettle: You would be correct: https://github.com/SUSE/doc-ses/blob/master/xml/entity-decl.ent#L23

xml/admin_ceph_upgrade.xml
lproven: OK then. Objection withdrawn and may @cwickert have mercy on our souls.

xml/admin_ceph_upgrade.xml
swiftgist: Is this question the context?  As in, this is the original SES5 content and the backend was new at the time, but for SES6 this is no longer new.

With respect to referencing Bluestore as the default backend for OSDs, is it better to follow the product, project or not mention it.  I can see the value in mentioning it if part of the goal is to persuade the reader to migrating.

xml/admin_ceph_upgrade.xml
swiftgist: Are you thinking that some version of SES6 did not support bluestore and filestore?

xml/admin_ceph_upgrade.xml
swiftgist: So, this is where I explain things but my intended audience in the conversation of this PR is the doc team.  I will go into additional details to help give enough information.  I cannot say that my explanation is perfectly suitable for the administrator reading the SES6 docs.  It's fine to cull what I said so that the manual comes from a single voice.

That said, I do somewhat expect that the technical reader of the SES6 manual would know what idempotent means.  OTOH, I do not know how many translations we support for the manual. Having a description of wouldn't hurt, but since I'm emphasizing the negative maybe a warning that `rebuild.node` will *always* remove and recreate all OSDs is more appropriate.

@lproven Where did you see the contradiction?  That will likely mean there would be cause for confusion.  The SES5 migration orchestration is idempotent.  The SES6 rebuild runner is not.

xml/admin_ceph_upgrade.xml
swiftgist: Do you wish enumerate every possible failure for a disk to be unformatted and then reformatted?  Beyond the type of failure, what is your expectation of the administrator reading this manual?  

If the `rebuild.node` fails to remove the second OSD, the administrator will need to investigate and correct the issue.  In this case, rerunning the `rebuild.node` for that minion makes the most sense because the already removed OSDs cannot be more removed.

If the rebuild.node removes all the OSDs, creates the first OSD, but fails to create the second OSD, the administrator must still intervene and correct the issue.  Once complete, the best course forward may be to rerun the `rebuild.node` again.  However, the first successfully created OSD will be destroyed and recreated.  Obviously, there is lost time here.  The other choice is to forgo using the `rebuild.node`, realize that you effectively have a new storage node with a single OSD deployed and revert to the section of adding a new storage node.  (This amounts to calling `salt-run disks.deploy target=TARGET`.)  

If half of the OSDs on the storage node had been created successfully, then I would not recommend rerunning `rebuild.node`.  Exceptions exist though such as a PoC with a trivial amount of data.  

How do you wish to convey this?

xml/admin_ceph_upgrade.xml
tbazant:
```suggestion
   option since &productname; 5. Compared to &filestore;, which
```

xml/admin_ceph_upgrade.xml
```diff
@@ -1397,6 +1397,146 @@ data:
...
+
+ <sect1 xml:id="filestore2bluestore">
+  <title>OSD Migration to &bluestore;</title>
+  <para>
+   OSD &bluestore; is a new back end for the OSD daemons. It is the default
+   option since &productname; &productnumber;. Compared to FileStore, which
+   stores objects as files in an XFS file system, &bluestore; can deliver
+   increased performance because it stores objects directly on the underlying
+   block device. &bluestore; also enables other features, such as built-in
+   compression and EC overwrites, that are unavailable with FileStore.
```
tbazant:
```suggestion
   compression and EC overwrites, that are unavailable with &filestore;.
```

xml/admin_ceph_upgrade.xml
tbazant:
```suggestion
   In &productname; 5, both &filestore; and &bluestore; are supported and it is possible
```

xml/admin_ceph_upgrade.xml
```diff
@@ -1397,6 +1397,146 @@ data:
...
+  <para>
+   Specifically for &bluestore;, an OSD has a 'wal' (Write Ahead Log) device
+   and a 'db' (RocksDB database) device. The RocksDB database holds the
+   metadata for a &bluestore; OSD. These two devices will reside on the same
+   device as an OSD by default, but either can be placed on faster/different
+   media.
+  </para>
+  <para>
+   In SES5, both FileStore and &bluestore; are supported and it is possible
+   for FileStore and &bluestore; OSDs to co-exist in a single cluster. During
```
tbazant:
```suggestion
   for &filestore; and &bluestore; OSDs to co-exist in a single cluster. During
```

xml/admin_ceph_upgrade.xml
tbazant:
```suggestion
   the &productname; upgrade procedure, &filestore; OSDs are not
```

xml/admin_ceph_upgrade.xml
```diff
@@ -1397,6 +1397,146 @@ data:
...
+   Before converting to &bluestore;, the OSDs need to be running &productname;
+   &productnumber;. The conversion is a slow process as all data gets
+   re-written twice. Though the migration process can take a long time to
+   complete, there is no cluster outage and all clients can continue accessing
+   the cluster during this period. However, do expect lower performance for
+   the duration of the migration. This is caused by rebalancing and
+   backfilling of cluster data.
+  </para>
+  <para>
+   Use the following procedure to migrate FileStore OSDs to &bluestore;:
```
tbazant:
```suggestion
   Use the following procedure to migrate &filestore; OSDs to &bluestore;:
```

xml/admin_ceph_upgrade.xml
tbazant: I remember that saying too...ancient times now, myeb true, but:
In single line &lt;screens>'s, it's quite ok to have it inline. But with multi-line ones, it is MUCH easier to edit the inner &lt;screen>'s content if it is like this. And honestly, it does not look bad either.

xml/admin_ceph_upgrade.xml
```diff
@@ -1397,6 +1397,146 @@ data:
...
+ <screen>
+ &prompt.smaster;salt-run disengage.safety
+ </screen>
+   <para>Rebuild the nodes before continuing:</para>
+ <screen>
+ &prompt.smaster; salt-run rebuild.node <replaceable>TARGET</replaceable>
+ </screen>
+   <para>You can also choose to rebuild each node individually. For example:</para>
+ <screen>
+   &prompt.smaster; salt-run rebuild.node data1.ceph
```
tbazant:
```suggestion
&prompt.smaster; salt-run rebuild.node data1.ceph
```

xml/admin_ceph_upgrade.xml
```diff
@@ -1397,6 +1397,146 @@ data:
...
+ </screen>
+   <para>The rebuild is not idempotent. To address the complication of shared
+     devices, the first step of a rebuild is to remove all the OSDs on a node.
+     If one OSD fails to convert, rerunning the rebuild will destroy the
+     already converted Bluestore OSDs. Depending on the failure and the
+     amount of progress, for example we recommend running:</para>
+ <screen>
+   &prompt.smaster; salt-run disks.deploy target=data1.ceph
```
tbazant:
```suggestion
&prompt.smaster; salt-run disks.deploy target=data1.ceph
```

xml/admin_ceph_upgrade.xml
```diff
@@ -1397,6 +1397,147 @@ data:
...
+ </screen>
+   <para>You can also choose to rebuild each node individually. For example:</para>
+ <screen>
+&prompt.smaster; salt-run rebuild.node data1.ceph
+ </screen>
+   <para>The <literal>rebuild.node</literal> always removes and recreates
+     all OSDs on the node.</para>
+     <important>
+       <para>If one OSD fails to convert, re-running the rebuild destroys the
+         already converted &bluestore; OSDs. If this happens, run:</para>
```
swiftgist: Is "If this happens" too terse?  Maybe "Instead of re-running the rebuild, you may run:"

xml/admin_ceph_upgrade.xml
```diff
@@ -1397,6 +1397,147 @@ data:
...
+     original data structure, and converts it to the new data structure. The
+     result is a new hardware profile named
+     'migrated-<replaceable>ORIGINAL_NAME</replaceable>'.
+     <filename>policy.cfg</filename> is updated as well.
+    </para>
+    <para>
+     If the original configuration had separate journals, the &bluestore;
+     configuration will use the same device for the 'wal' and 'db' for that
+     OSD.
+    </para>
```
swiftgist: I just double-checked.  While the state `ceph.migrate.policy` is there, the referenced module call no longer exists in SES6.  

This paragraph can be removed.

xml/admin_ceph_upgrade.xml
```diff
@@ -1397,6 +1397,147 @@ data:
...
+     </para>
+    </tip>
+   </step>
+  </procedure>
+  <para>
+   After the migration to &bluestore;, the object count will remain the same
+   and disk usage will be nearly the same.
+  </para>
+</sect1>
+
```
swiftgist: Likewise, the states `ceph.migrate.nodes` and `ceph.migrate.osds` exist, but the `osd.redeploy` module is absent in SES6.

These paragraphs can be removed as well.

PR #121, opened 2019-10-24
Updating swift interface access to include SLE 15 SP1 (BSC#1155009)

asettle:


xml/admin_ceph_gateway.xml
```diff
@@ -288,12 +288,16 @@ for bucket in conn.get_all_buckets():
      you more about its command line options.
     </para>
     <para>
-     The package is included in the 'Public Cloud' module for &sle; 12 SP3.
+     The package is included in the 'Public Cloud' module for &sle; 12 SP3 and
+     &sle; 15 SP1.
```
ajaeger: since 12 SP3 (it's in SP4, SP5 as well) - and we can generalize for 15, it's in GA, SP2, what about just saying "SLES 12 since SP3 and SLES 15" ?

xml/admin_ceph_gateway.xml
```diff
@@ -288,12 +288,16 @@ for bucket in conn.get_all_buckets():
...
      Before installing the package, you need to activate the module and refresh
      the software repository:
     </para>
 <screen>&prompt.root;SUSEConnect -p sle-module-public-cloud/12/x86_64
```
ajaeger: Want to update for ARM here as well?

asettle: We don't have thorough, consistent, documentation for ARM. Whilst I agree we should be including ARM in locations like this, I think that's a separate issue that we should be doing in one PR, rather than in small nits like this one.
I am happy to create another bug to ensure we go through the entire suite and look for ARM-related updates.

ajaeger: Ok, let's do separately

xml/admin_ceph_gateway.xml
```diff
@@ -288,12 +288,16 @@ for bucket in conn.get_all_buckets():
...
+     The package is included in the 'Public Cloud' module for &sle; 12 since SP3 and
+     &sle; 15.
      Before installing the package, you need to activate the module and refresh
      the software repository:
     </para>
 <screen>&prompt.root;SUSEConnect -p sle-module-public-cloud/12/x86_64
 sudo zypper refresh</screen>
+     <para>Or</para>
+<screen>&prompt.root;SUSEConnect -p sle-module-public-cloud/15/x86_64
+sudo zypper refresh</screen>
```
tbazant: This looks inconsistent to me, myabe just a heritage, but: in one &lt;screent> element, you start with root prompt, while on its 2nd line, you use `sudo`
```suggestion
&prompt.root;zypper refresh</screen>     
```

PR #127, opened 2019-11-01
Run 'rpmconfigcheck' after each node upgrade (jsc#SES-348)

tbazant:
It is advisable to run `rpmconfigcheck` after each node upgrade to verify that no local changes
to  configuration files are lost.


tbazant:
@jan--f you are right, the additional changes are caused by inconsistent XML-formatting - they do not introduce/modify the content

PR #134, opened 2019-11-14
Made network recommendations more specific (bsc#1156631)

tbazant:
The general paragraph about network recommendations was a bit vague and did not refer
to a SLE section which explains how to set bonding devices. This PR amends that.

### DIFF
--- a/xml/deployment_hwrecommend.xml
+++ b/xml/deployment_hwrecommend.xml
@@ -375,12 +375,11 @@
   <title>Network Recommendations</title>
 
   <para>
-   For the &ceph; cluster network environment, we recommend either two bonded
-   10 GbE network interfaces, or one 25 GbE network interface. The network
-   interfaces should be split into a public client part, and a trusted internal
-   part using VLANs. Use the 802.3ad bonding mode to provide maximum bandwidth
-   and resiliency. Find more details about configuring bonded network
-   interfaces in
+   For the &ceph; cluster network environment, we recommend two bonded 25 GbE
+   network interfaces. The network interfaces should be split into a public
+   client part, and a trusted internal part using VLANs. Use the 802.3ad
+   bonding mode to provide maximum bandwidth and resiliency. Find more details
+   about configuring bonded network interfaces in
    <link xlink:href="https://documentation.suse.com/sles/15-SP1/html/SLES-all/cha-network.html#sec-network-iface-bonding"/>
   </para>
 
--- a/xml/deployment_hwrecommend.xml
+++ b/xml/deployment_hwrecommend.xml
@@ -375,11 +375,12 @@
   <title>Network Recommendations</title>
 
   <para>
-   The network environment where you intend to run &ceph; should ideally be a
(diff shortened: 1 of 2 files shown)


xml/deployment_hwrecommend.xml
```diff
@@ -375,11 +375,12 @@
   <title>Network Recommendations</title>
 
   <para>
-   The network environment where you intend to run &ceph; should ideally be a
-   bonded set of at least two network interfaces that is logically split into a
-   public part and a trusted internal part using VLANs. The bonding mode is
-   recommended to be 802.3ad if possible to provide maximum bandwidth and
-   resiliency.
+   For the &ceph; cluster network environment, we recommend a bonded set of at
+   least two separate 25 GbE network interfaces. The networks interfaces should
```
ajaeger: two bonded 10 GbE or one 25 GbE AFAIK

xml/deployment_hwrecommend.xml
```diff
@@ -375,11 +375,13 @@
   <title>Network Recommendations</title>
 
   <para>
-   The network environment where you intend to run &ceph; should ideally be a
-   bonded set of at least two network interfaces that is logically split into a
-   public part and a trusted internal part using VLANs. The bonding mode is
-   recommended to be 802.3ad if possible to provide maximum bandwidth and
-   resiliency.
+   For the &ceph; cluster network environment, we recommend either two bonded
+   10 GbE network interfaces, or one 25 GbE network interface. The network
```
capri1989: Is that our recommendation? I would start with at least 25GbE bonded network and given the price of 100GbE in the meanwhile everyone should aim for that in a "larger" enviroment and also if we talk about all-flash in the meanwhile. 

Someone could of course also use 10GbE but in 2019 I think we should at least start with 25GbE not only because of the bandwith but foremost because of the latency.

ajaeger: You're right, we " required 4x 10 GbE, recommended 2x 25 GbE" - see https://documentation.suse.com/ses/6/single-html/ses-deployment/#sysreq-osd. Going with 100 GbE is a good recommendation but needs more work overall.

xml/deployment_hwrecommend.xml
tbazant: @ajaeger so.....how far did we get with the NIC specifications? any last suggestion?

ajaeger: Might take a bit longer to get this done...

tbazant: ok, when should i ping you again, in a week?

xml/deployment_hwrecommend.xml
asettle: @tbazant - Kai has confirmed this is accurate.

xml/deployment_hwrecommend.xml
asettle: @tbazant - Kai's response below answers this query.

PR #137, opened 2019-11-21
added a new table with description of columns

tbazant:


xml/admin_operating_pools.xml
```diff
@@ -304,16 +304,139 @@ mon_allow_pool_delete config option to true before you can destroy a pool</scree
...
+      </para>
+     </listitem>
+    </varlistentry>
+    <varlistentry>
+     <term>CLONES</term>
+     <listitem>
+      <para>
+       Number of clones stored in the pool. When a snapshot is created and one
+       writes to an object, instead of modifying the original object its clone
+       is created so the original 'snapshotted' object content is not modified.
```
asettle:
```suggestion
       is created so the original snapshotted object content is not modified.

File: xml/admin_operating_pools.xml [hunk 2]
```diff
@@ -304,16 +304,139 @@ mon_allow_pool_delete config option to true before you can destroy a pool</scree
...
+      </para>
+     </listitem>
+    </varlistentry>
+    <varlistentry>
+     <term>COPIES</term>
+     <listitem>
+      <para>
+       Number of object replicas. For example if a replicated pool with the
```
asettle:
```suggestion
       Number of object replicas. For example, if a replicated pool with the

PR #138, opened 2019-11-22
Section about @/var/lib/ceph subvolume now covers all scenarios (bsc#1155450)

tbazant:
Skipping the creation of @/var/lib/ceph subvolume causes failure when running DeepSea stage.3
This addition covers all scenarios that might happen during the cluster deployment/upgrade.


xml/admin_ceph_tips.xml
```diff
@@ -367,8 +367,83 @@ osd_scrub_load_threshold = 0.25
...
+     <listitem>
+      <para>
+       Creates <filename>/var/lib/ceph</filename> as a
+       <literal>@/var/lib/ceph</literal> Btrfs subvolume.
+      </para>
+     </listitem>
+     <listitem>
+      <para>
+       Mount the new subvolume and update <filename>/etc/fstab</filename>
+       appropriately.
```
swiftgist: Shouldn't this be "Mounts" and "updates" to match "Creates"?

swiftgist:
I made the one comment, otherwise, lgtm. :)

xml/admin_ceph_tips.xml
tbazant: yep, fixed

PR #142, opened 2019-12-03
Adding additional information on non-SUSE OpenStack distros (BSC#1158222)

asettle:


asettle:
> I think a inside a is not permitted by GeekoDoc.

Aye, it's fixed already.

PR #146, opened 2019-12-09
Fill in values for "pod", "region" & "root" in bucket list. (BSC#1113055)

lproven:
I found definitions of the terms here:
https://docs.ceph.com/docs/master/rados/operations/crush-map/

I found some standard geographical region abbreviations here:
https://www.seforall.org/sites/default/files/l/2013/09/5-gtf_overview.pdf

Using this info & lacking any other feedback, I have filled in the missing boxes.

Oxygen has helpfully reflowed the file for me -- I have attempted to correct this with DAPS-xmlformat.

### DIFF
--- a/xml/admin_ceph_datamgm.xml
+++ b/xml/admin_ceph_datamgm.xml
@@ -745,7 +745,7 @@ maps appear equivalent
       </entry>
       <entry>
        <para>
-        A logical geographical area containing one of more zones (e.g. NAM, LAM,
+        A logical geographical area containing one or more zones (e.g. NAM, LAM,
         EMEA, APAC etc.)
        </para>
       </entry>
--- a/xml/admin_ceph_datamgm.xml
+++ b/xml/admin_ceph_datamgm.xml
@@ -584,7 +584,8 @@ maps appear equivalent
       </entry>
       <entry>
        <para>
-        An OSD daemon (osd.1, osd.2, etc.).
+        A specific &ceph; Object Storage Daemon (<literal>osd.1</literal>,
+        <literal>osd.2</literal>, etc.).
        </para>
       </entry>
      </row>
@@ -601,7 +602,7 @@ maps appear equivalent
       </entry>
       <entry>
        <para>
-        A host name containing one or more OSDs.
+        The name of a host containing one or more OSDs.
        </para>
       </entry>
      </row>
@@ -618,7 +619,8 @@ maps appear equivalent
       </entry>
       <entry>
        <para>
-        Chassis of which the rack is composed.
+        Identifier for which chassis in the rack contains the
+        <literal>host</literal>.
        </para>
       </entry>
      </row>
@@ -669,7 +671,7 @@ maps appear equivalent
       </entry>
       <entry>
        <para>
-        Power Distribution Unit (abbreviation).
+        Abbreviation for Power Distribution Unit.
        </para>
       </entry>
      </row>
@@ -720,7 +722,7 @@ maps appear equivalent
       </entry>
       <entry>
        <para>
-        A physical data center containing rooms.
+        A physical data center containing one or more rooms.
        </para>
       </entry>
      </row>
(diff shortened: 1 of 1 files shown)


xml/admin_ceph_datamgm.xml
```diff
@@ -653,7 +669,7 @@ maps appear equivalent
       </entry>
       <entry>
        <para>
-        Power distribution unit.
+        Power Distribution Unit (abbreviation).
```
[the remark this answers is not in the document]
lproven: Fair call -- that *is* ambiguous. I will reword.

xml/admin_ceph_datamgm.xml
```diff
@@ -67,31 +70,36 @@
  <para>
   A custom &crushmap; can also help you identify the physical locations where
   &ceph; stores redundant copies of data when the placement group(s) (refer to
-  <xref linkend="op-pgs"/>) associated with a failed host are in a degraded
-  state.
+  <xref
```
tbazant: Not sure that splitting xref across 3 lines is useful :-)

lproven: That was daps-xmlformat, I fear. :(

lproven:
> Sorry for being late to the party. Two comments:
> The hierarchy should be
> 
>     1. osd
>     2. host
>     3. chassis
>     4. rack
>     5. pdu
>     6. pod
>     7. row
>     8. room
>     9. datacenter
>     10. region
>     11. root

That conflicts with the Ceph docs, here:

https://docs.ceph.com/docs/master/rados/operations/crush-map/#types-and-buckets

You have
rack < pdu < pod < row

They have: 
rack < row < pdu < pod

This implies that a pod might have >1 PDUs, which seems reasonable, so I will reword that.

> An hence a pod would be a group of racks.

Agreed! However you have caused me to notice that we are missing "zone" from the list, between "datacenter" and "region".

> Otherwise looks good +1

Thank you!

jan--f:
> That conflicts with the Ceph docs, here:
> 
> https://docs.ceph.com/docs/master/rados/operations/crush-map/#types-and-buckets
> 
> You have
> rack < pdu < pod < row
> 
> They have:
> rack < row < pdu < pod

Hmm I was looking at https://docs.ceph.com/docs/master/rados/operations/crush-map/#crush-location point 2.

I think one pdu is unlikely to power multiple rows, but also far from being an expert here. Maybe @l-mb can shed additional light?

xml/admin_ceph_datamgm.xml
```diff
@@ -727,13 +738,34 @@ maps appear equivalent
         10
        </para>
       </entry>
+      <entry>
+       <para>
+        region
```
theanalyst: Region is no longer used and deprecated in the cli tools as well wrt RGW, the term used is Zonegroups which are a collection of Zones. For clusers older than SES3 this corresponds to Regions which was the older term

lproven: > Region is no longer used and deprecated in the cli tools as well wrt RGW,

Could you explain what "as well wrt RGW" means, please? Because it is so terse I do not follow. Does it mean that the term "RGW" is deprecated? 

> the term used is Zonegroups which are a collection of Zones. For clusers older than SES3 this corresponds to Regions which was the older term

These are the SES 6 docs, currently. That is all we are considering right now in this change.

theanalyst: So regions were the old term for a collection of Zones, since this wasn't exactly mapping the aws concept of a region, these were renamed to Zonegroups. the radosgw admin cli tools and everything else also don't use the region term anymore (https://docs.ceph.com/docs/master/radosgw/multisite/ has a couple of words on this)

lproven: Please bear in mind that this is solely relating to CRUSH map buckets. I do not work with Ceph, AWS or anything related to it myself, I just help maintain the docs, so I cannot really follow explanations in such terms.

I have reworded it thus:

«
10 | zonegroup | A group of zones. Replaces the older geographically-based "region".
»

Is that better?

theanalyst: hm I actually misread this section, I assumed this was a general appendix glossary, purely in crush context I don;t think zones even exist, region could just be a geographical region containing datacenters

PR #153, opened 2020-01-14
'salt osd.retain' is obsoleted (bsc#1160219)

tbazant:
The 'salt osd.retain' command for updating Salt grains after OSD recovery is obsoleted in SES6. This update suggest alternative way to remove stale grains.

### DIFF
--- a/xml/admin_saltcluster.xml
+++ b/xml/admin_saltcluster.xml
@@ -785,7 +785,7 @@ Device Path Size rotates available Model name
    </step>
    <step>
     <para>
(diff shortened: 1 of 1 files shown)


xml/admin_saltcluster.xml
```diff
@@ -785,9 +785,9 @@ Device Path Size rotates available Model name
    </step>
    <step>
     <para>
-     Update &salt; grains:
+     Remove unneeded &salt; grains (best after all OSDs have been migrated to LVM):
```
asettle:
```suggestion
     Remove unnecessary &salt; grains (best after all OSDs have been migrated to LVM):

PR #156, opened 2020-01-15
Removing specific instances of x86 in favor of <replaceable> (BSC#1155012)

asettle:
SES 6 supports both x86 and arm arch, however a lot of commands
specifically poin to x86. This patch offers a suggestion on how to
change this.


ajaeger:
LGTM, thanks.

PR #164, opened 2020-01-29
Added warning about adjusting pgp_num along with pg_num (BSC #1161690)

lproven:
Fixes:
https://bugzilla.suse.com/show_bug.cgi?id=1161690

As requested by @Martin-Weiss


lproven:
> i suggest using <option>'s for these <literal>'s but...

Fixed.

PR #165, opened 2020-01-29
Added pool migration limitations (bsc#1140080)

tbazant:
This update aligns an existing section on pool migration and prepends the subsections with general limitation of mentioned migration methods.

### DIFF
--- a/xml/admin_operating_pools.xml
+++ b/xml/admin_operating_pools.xml
@@ -910,8 +910,11 @@ pool4                     1.1 MiB      13      0     39                  0      
     <listitem>
      <para>
       You cannot migrate RBD images and &cephfs; exports from a replicated pool
-      to an EC pool. EC pools can store data but not metadata. The header
-      object of the RBD will fail to be flushed. The same applies for &cephfs;.
+      to an EC pool. The reason is that EC pools do not support
+      <literal>omap</literal>, while RBD and &cephfs; use
+      <literal>omap</literal> to store its metadata. For example, the header
+      object of the RBD will fail to be flushed. But you can migrate data to EC
+      pool, leaving metadata in replicated pool.
      </para>
     </listitem>
     <listitem>
--- a/xml/admin_operating_pools.xml
+++ b/xml/admin_operating_pools.xml
@@ -923,7 +923,7 @@ pool4                     1.1 MiB      13      0     39                  0      
       Nautilus or newer) will be able to open the image just after the
       'prepare' step, while older <systemitem>librbd</systemitem> clients or
       the <systemitem>krbd</systemitem> clients will not be able to open the
-      image until the 'commit' step is executed.
+      image until the <option>commit</option> step is executed.
      </para>
     </listitem>
    </itemizedlist>
--- a/xml/admin_operating_pools.xml
+++ b/xml/admin_operating_pools.xml
@@ -918,7 +918,7 @@ pool4                     1.1 MiB      13      0     39                  0      
      <para>
       The <command>rbd migration</command> method allows migrating images with
       minimal client downtime. You only need to stop the client before the
-      'prepare' step and start it afterward. Note that only a
+      <option>prepare</option> step and start it afterward. Note that only a
       <systemitem>librbd</systemitem> client that supports this feature (&ceph;
       Nautilus or newer) will be able to open the image just after the
       'prepare' step, while older <systemitem>librbd</systemitem> clients or
--- a/xml/admin_operating_pools.xml
+++ b/xml/admin_operating_pools.xml
@@ -891,7 +891,7 @@ pool4                     1.1 MiB      13      0     39                  0      
    deployment.
   </para>
   <para>
-   There are two migration methods&mdash;a <emphasis>cache tier</emphasis>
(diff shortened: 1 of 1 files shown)


xml/admin_operating_pools.xml
```diff
@@ -890,23 +890,47 @@ pool4                     1.1 MiB      13      0     39                  0
...
+      You cannot migrate RBD images and &cephfs; exports from a replicated pool
+      to an EC pool. EC pools can store data but not metadata. The header
+      object of the RBD will fail to be flushed. The same applies for &cephfs;.
+     </para>
+    </listitem>
+    <listitem>
+     <para>
+      The <command>rbd migration</command> method allows migrating images with
+      minimal client downtime. You only need to stop the client before the
+      'prepare' step and start it afterward. Note that only a
```
asettle:
```suggestion
      <option>prepare</option> step and start it afterward. Note that only a

File: xml/admin_operating_pools.xml [hunk 2]
```diff
@@ -890,23 +890,47 @@ pool4                     1.1 MiB      13      0     39                  0
...
+      <systemitem>librbd</systemitem> client that supports this feature (&ceph;
+      Nautilus or newer) will be able to open the image just after the
+      'prepare' step, while older <systemitem>librbd</systemitem> clients or
```
asettle:
```suggestion
      <option>prepare</option> step, while older <systemitem>librbd</systemitem> clients or

File: xml/admin_operating_pools.xml [hunk 3]
```diff
@@ -890,23 +890,47 @@ pool4                     1.1 MiB      13      0     39                  0
...
+      the <systemitem>krbd</systemitem> clients will not be able to open the
+      image until the 'commit' step is executed.
```
asettle:
```suggestion
      image until the <option>commit</option> step is executed.

**Comment by tbazant** [comment]
Adding a rendered version
[pools-migration_color_en.pdf](https://github.com/SUSE/doc-ses/files/4128411/pools-migration_color_en.pdf)

File: xml/admin_operating_pools.xml [hunk 4]
```diff
@@ -890,23 +890,47 @@ pool4                     1.1 MiB      13      0     39                  0
...
+    migrate</command> sub-commands to migrate RBD images to a new pool. Each
+   method has its specifics and limitations.
   </para>
 
+  <sect2 xml:id="pool-migrate-limits">
+   <title>Limitations</title>
+   <itemizedlist>
+    <listitem>
+     <para>
+      You can use the cache tier method to migrate from a replicated pool to
```
lproven: Should this not be highlighted/emphasized  as the `rbd migration` method is, below?

xml/admin_operating_pools.xml
```diff
@@ -1045,18 +1069,10 @@ pool4                     1.1 MiB      13      0     39                  0
...
-     of the RBD will fail to be flushed. The same applies for &cephfs;.
-    </para>
-   </warning>
   </sect2>
 
   <sect2 xml:id="migrate-rbd-image">
-   <title>Migrating a &rbd; Image</title>
+   <title>Migrating RBD Images</title>
    <para>
     Following is the recommended way to migrate RBD images from one replicated
```
lproven:
```suggestion
    The following is the recommended way to migrate RBD images from one replicated
```

xml/admin_operating_pools.xml
```diff
@@ -890,23 +890,47 @@ pool4                     1.1 MiB      13      0     39                  0
    need to migrate the pool data to another one whose parameters suit your
    deployment.
   </para>
-
   <para>
-   There are several methods of pool migration. We recommend using
-   <emphasis>cache tier</emphasis> because this method is transparent, reduces
-   cluster downtime, and avoids duplicating the entire pool's data.
+   There are two migration methods&mdash;a <emphasis>cache tier</emphasis>
```
trociny: I think "there are" more than two methods. But only two of them are described here. So, probably change to something like "Thi document describes two migration methods ..."?

xml/admin_operating_pools.xml
```diff
@@ -890,23 +890,47 @@ pool4                     1.1 MiB      13      0     39                  0
...
+      You can use the cache tier method to migrate from a replicated pool to
+      either an EC pool or another replicated pool. Migrating from an
+      EC pool is not supported.
+     </para>
+    </listitem>
+    <listitem>
+     <para>
+      You cannot migrate RBD images and &cephfs; exports from a replicated pool
+      to an EC pool. EC pools can store data but not metadata. The header
+      object of the RBD will fail to be flushed. The same applies for &cephfs;.
```
trociny: Actually it is more than just the header object that will fail to migrate. There are other rbd objects that contain metadata (omap data) that will also fail to migrate. So I would change it to: "For example, the header object of the RBD will faill ...", or just skip this part.

Also, it probably worth mentioning that by metadata we mean any data stored as omap key-values, and EC does not support omap. So, instead of "EC pools can store data but not metadata." I would prefer something like: "Erasure coded pools do not support omap,  and RBD and cephfs use omap to store its metadata. But you can migrate data to EC pool, leaving metadata in replicated pool".

PR #167, opened 2020-01-29
Add warning about free space when adding replicas (bsc#1161691)

lproven:
After discussion with @Martin-Weiss, added a warning about free cluster space when adding replicas. 

This fixes:
https://bugzilla.suse.com/show_bug.cgi?id=1161691

### DIFF
--- a/xml/ceph_maintenance_updates.xml
+++ b/xml/ceph_maintenance_updates.xml
@@ -92,7 +92,7 @@
     <title>Additional Replicas Require Extra Space</title>
     <para>
      Ensure there is sufficent free space <emphasis>before</emphasis> adding
-     replicas. You can check using the commands: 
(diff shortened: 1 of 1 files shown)


xml/ceph_maintenance_updates.xml
```diff
@@ -88,6 +88,25 @@
 <screen>
 &prompt.cephuser;ceph config set global mon_warn_on_pool_no_redundancy false
 </screen>
+   <warning>
+    <title>Additional Replicas Require Extra Space</title>
+    <para>
+     Ensure there is sufficent free space <emphasis>before</emphasis> adding
+     replicas. You can check using the commands:
```
asettle:
```suggestion
     replicas. You can check using the following commands:

PR #168, opened 2020-01-30
Shutdown and start of the whole cluster (bsc#1126210)

tbazant:
Updated and extended the way and order of nodes during a complete cluster shutdown/start-up.
This update applies to SES6+

### DIFF
--- a/xml/admin_operating_services.xml
+++ b/xml/admin_operating_services.xml
@@ -150,7 +150,7 @@ ceph-rbd-mirror@.service
   </sect2>
  </sect1>
  <sect1 xml:id="Deepsea-restart">
-  <title>Restarting &ceph; Services USing &deepsea;</title>
+  <title>Restarting &ceph; Services Using &deepsea;</title>
 
   <para>
    After applying updates to the cluster nodes, the affected &ceph; related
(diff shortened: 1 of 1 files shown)


tbazant:
Adding a rendered version.
[ceph-cluster-shutdown_color_en.pdf](https://github.com/SUSE/doc-ses/files/4134724/ceph-cluster-shutdown_color_en.pdf)

xml/admin_operating_services.xml
```diff
@@ -244,31 +244,78 @@ ceph-rbd-mirror@.service
...
+  <procedure>
+   <title>Shutting Down the Whole &ceph; Cluster</title>
+   <step>
+    <para>
+     Shut down or disconnect any clients accessing the cluster.
+    </para>
+   </step>
+   <step>
+    <para>
+     To prevent CRUSH to automatically rebalance the cluster, set the cluster
```
asettle:
```suggestion
     To prevent CRUSH automatically rebalancing the cluster, set the cluster

Comment (tbazant) [reply xml/admin_operating_services.xml]: copy pasta error...but should this not be "prevent CRUSH from automatically..."?

Comment (asettle) [reply xml/admin_operating_services.xml]: @tbazant either or. "from" is nicer :)

Comment (asettle) [reply xml/admin_operating_services.xml]: I've edited it so you can just add the suggestion :D

Comment (tbazant) [inline xml/admin_operating_services.xml]:
```suggestion
     To prevent CRUSH from automatically rebalancing the cluster, set the cluster
```

PR #169, opened 2020-01-31
Adding new note defining log action (BSC#1157612)

asettle:


denisok:
example is missed
`
[client]
log_file = "/var/log/ceph/ceph-client.log"`

PR #171, opened 2020-02-03
Shutdown and starting of whole Ceph cluster (bsc#1126210)

tbazant:
This update describes detailed steps to shut down and start all nodes of the Ceph cluster.

### DIFF
--- a/xml/admin_operating_services.xml
+++ b/xml/admin_operating_services.xml
@@ -238,7 +238,7 @@ ceph-rbd-mirror@.service
 
   <para>
    There are occasions when you need to stop all &ceph; related services in the
-   cluster in the recommended order, and then be able to simply start them
+   cluster in the recommended order and restart without issue.
    again. For example, in case of a planned power outage.
   </para>
 
--- a/xml/admin_operating_services.xml
+++ b/xml/admin_operating_services.xml
@@ -233,4 +233,127 @@ ceph-rbd-mirror@.service
 <screen>&prompt.smaster;salt-run state.orch ceph.restart.ganesha</screen>
   </sect2>
  </sect1>
+ <sect1 xml:id="ceph-cluster-shutdown">
+  <title>Shutdown and Start of the Whole &ceph; Cluster</title>
+
+  <para>
+   There are occasions when you need to stop all &ceph; related services in the
+   cluster in the recommended order, and then be able to simply start them
+   again. For example, in case of a planned power outage.
+  </para>
+
+  <procedure>
+   <title>Shutting Down the Whole &ceph; Cluster</title>
+   <step>
+    <para>
+     Shut down or disconnect any clients accessing the cluster.
+    </para>
+   </step>
+   <step>
+    <para>
+     To prevent CRUSH from automatically rebalancing the cluster, set the
+     cluster to <literal>noout</literal>:
+    </para>
+<screen>&prompt.smaster;ceph osd set noout</screen>
+   </step>
+   <step>
+    <para>
+     Disable safety measures:
+    </para>
+<screen>
+&prompt.smaster;salt-run disengage.safety
+</screen>
+   </step>
+   <step>
+    <para>
+     Stop all &ceph; services in the following order:
+    </para>
+    <substeps>
+     <step>
+      <para>
+       Stop &ganesha;:
+      </para>
+<screen>&prompt.smaster;salt -C 'I@roles:ganesha and I@cluster:ceph' ceph.terminate.ganesha</screen>
+     </step>
+     <step>
+      <para>
+       Stop &ogw;s:
+      </para>
+<screen>&prompt.smaster;salt -C 'I@roles:rgw and I@cluster:ceph' ceph.terminate.rgw</screen>
+     </step>
+     <step>
+      <para>
+       Stop &mds;s:
+      </para>
+<screen>&prompt.smaster;salt -C 'I@roles:mds and I@cluster:ceph' ceph.terminate.mds</screen>
+     </step>
+     <step>
+      <para>
+       Stop &igw;s:
+      </para>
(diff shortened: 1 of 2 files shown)


xml/admin_operating_services.xml
```diff
@@ -233,4 +233,127 @@ ceph-rbd-mirror@.service
 <screen>&prompt.smaster;salt-run state.orch ceph.restart.ganesha</screen>
   </sect2>
  </sect1>
+ <sect1 xml:id="ceph-cluster-shutdown">
+  <title>Shutdown and Start of the Whole &ceph; Cluster</title>
+
+  <para>
+   There are occasions when you need to stop all &ceph; related services in the
+   cluster in the recommended order, and then be able to simply start them
```
asettle:
```suggestion
   cluster in the recommended order and restart without issue.

File: xml/admin_operating_services.xml [hunk 2]
```diff
@@ -233,4 +233,127 @@ ceph-rbd-mirror@.service
...
+   again. For example, in case of a planned power outage.
+  </para>
+
+  <procedure>
+   <title>Shutting Down the Whole &ceph; Cluster</title>
```
asettle: Probably don't need this title, since it's the title of the section too

xml/admin_operating_services.xml
```diff
@@ -233,4 +233,127 @@ ceph-rbd-mirror@.service
...
+   <step>
+    <para>
+     Shut down or disconnect any clients accessing the cluster.
```
asettle: Do we want to provide CLI instruction?

xml/admin_operating_services.xml
```diff
@@ -233,4 +233,127 @@ ceph-rbd-mirror@.service
 <screen>&prompt.smaster;salt-run state.orch ceph.restart.ganesha</screen>
   </sect2>
  </sect1>
+ <sect1 xml:id="ceph-cluster-shutdown">
+  <title>Shutdown and Start of the Whole &ceph; Cluster</title>
+
+  <para>
+   There are occasions when you need to stop all &ceph; related services in the
+   cluster in the recommended order, and then be able to simply start them
+   again. For example, in case of a planned power outage.
```
swiftgist: I think the two lines might be better reversed.

```
Shutting down and starting the cluster may be necessary such as in the case of a planned power outage.
To stop all &ceph; related services and restart without issue, follow the steps below.
```

xml/admin_operating_services.xml
swiftgist: The problem here is that the clients will be significantly different and the steps will vary.  In the case of CephFS and Ganesha, unmount any filesystems from Ceph.  For iSCSI, the steps might involve disconnecting LUNs in a VMware GUI.  For RGW, any automation for S3 clients needs to be disabled.  

I don't think we will have generic instructions that would apply to a specific customer and I doubt my description would be that helpful.

tbazant: Agree

xml/admin_operating_services.xml
tbazant: I dont agree, there are 2 adjacent procedures and the reader should know which one does what, or no? One for shutdown and one for startup.

xml/admin_operating_services.xml
```diff
@@ -233,4 +233,127 @@ ceph-rbd-mirror@.service
 <screen>&prompt.smaster;salt-run state.orch ceph.restart.ganesha</screen>
   </sect2>
  </sect1>
+ <sect1 xml:id="ceph-cluster-shutdown">
+  <title>Shutdown and Start of the Whole &ceph; Cluster</title>
+
+  <para>
+   There are occasions when you need to stop all &ceph; related services in the
+   cluster in the recommended order and restart without issue.
+   again. For example, in case of a planned power outage.
```
tbazant:
```suggestion
For example, in case of a planned power outage.
```

xml/admin_operating_services.xml
asettle: Hmmm okay, you're right about that. Leave it.

xml/admin_operating_services.xml
```diff
@@ -233,4 +233,127 @@ ceph-rbd-mirror@.service
 <screen>&prompt.smaster;salt-run state.orch ceph.restart.ganesha</screen>
   </sect2>
  </sect1>
+ <sect1 xml:id="ceph-cluster-shutdown">
+  <title>Shutdown and Start of the Whole &ceph; Cluster</title>
```
lproven:
```suggestion
  <title>Shutdown and Restart of the Whole &ceph; Cluster</title>
```

xml/admin_operating_services.xml
```diff
@@ -233,4 +233,127 @@ ceph-rbd-mirror@.service
...
+
+  <para>
+   There are occasions when you need to stop all &ceph; related services in the
```
lproven: I think this would be better with a small reword.

The text after "for example" is a sentence fragment; therefore, I suggest removing the full stop before it and replacing with a comma or a semi-colon, or even an en dash.

The phrase about "stopping all the services in the correct order" -- this is a clarification. I would remove it from that sentence. You just want to shut down cleanly and safely, right? So what this section provides is instructions how to do it in the correct order.

I was about to add a suggestion but I see that @jan--f already has done so. I endorse that.

xml/admin_operating_services.xml
```diff
@@ -237,9 +237,9 @@ ceph-rbd-mirror@.service
   <title>Shutdown and Restart of the Whole &ceph; Cluster</title>
 
   <para>
-   There are occasions when you need to stop all &ceph; related services in the
-   cluster in the recommended order and restart without issue. For example, in
-   case of a planned power outage.
+   Shutting down and restarting the cluster may be necessary such as in the
```
asettle:
```suggestions
   Shutting down and restarting the cluster may be necessary in the

xml/admin_operating_services.xml
tbazant: fixed

PR #172, opened 2020-02-04
Adding more content for networking overview section (JSC#1116)

asettle:
Wording from David Byte


dmbyte:
Looks good to me.

asettle:
> It is better but TBH I still do not understand it.
> 
> If I am reading this correctly, you:
> [1] take _two_ NICs & bond them into a single interface, and then
> [2] somehow you split that single interface into 2 different interfaces again using VLANs?
> 
> I am not an expert in network switching but I do not understand how stage 2 relates to stage 1.

@dmbyte ?

xml/deployment_hwrecommend.xml
```diff
@@ -378,24 +378,30 @@
...
+    Enhance fault tolerance through isolating the components into failure domains.
+    To improve fault tolerance of the network, bonding one interface from two
+    separate Network Interface Cards (NIC) offers protection against failure
+    of a single NIC. Similarly, creating a bond across two switches protects
+    against failure of a switch. We recommend consulting with the network
+    equipment vendor in order to architect the level of fault tolerance required.
+  </para>
+  <para>
+    The cluster utilizes two networks. The cluster network is used exclusively
+    for private backend communication, while the public address is used for
```
matthewoliver: we talk about 2 networks. The cluster "network" and then here you use the public "address". I assume address here should be network too.

matthewoliver:
> It is better but TBH I still do not understand it.
> 
> If I am reading this correctly, you:
> [1] take _two_ NICs & bond them into a single interface, and then
> [2] somehow you split that single interface into 2 different interfaces again using VLANs?
> 
> I am not an expert in network switching but I do not understand how stage 2 relates to stage 1.

By taking 2 NICs and bonding them, you get 2x the bandwidth. Also their connected to different switches and a NIC cards so you get fault tollerance.
You still need to support 2 networks, so then you place 2 VLANs. why? Read on:

If you kept it as 2 seperate NICs for the 2 networks. then the max bandwidth for a network is 1 NIC. And if a NIC or Switch does down, you loose that network on the host.

On the other hand, if you bond them, then the max bandwidth for a network is 2xNics. Yes you have 2 networks sharing 2 nics, So maybe 2 busy networks could cause issuse because not as good isolation, but still I think the pros out weight the cons. Also we're probably talking about tuning for performance. And this maximises the bandwidth for user data. 
Further, if a NIC or Switch goes down, we only loose some bandwidth, non of the networks are down on the host. So a bit of performance degregation but no loss of the node (which would be much worse).
Also means you could potentially do things like upgrade network infrastructure while keeping nodes up.

Sure it may look confusing tho, bond 2 physical nics. Then, from what it looks like in linux interfaces, split the bond into 2 virtual vlan nics. but in essence it's just giving the same bond to 2 networks.

asettle:
> The changes here are fine - but I suggest to do more. See below.
> 
> Let's be consistent with terminology, can we use the upstream
> terminology "cluster network" and "public network"? If not, let's do
> it as proposed below
> 
> New section
> 
> 2.1 Network Overview
> 
> (Can we add a graph like my slide 3?)
> 
> Ceph has several logical networks:
> 
> * A trusted internal network, the back-end network. Ceph calls this
>   the "cluster network".
> * A public client network. Ceph calls this the "public network".
> * Client networks for gateways, these are optional
> 
> The trusted internal network is the back-end network between the OSD
> nodes for replication and balancing, it needs roughly three times the
> bandwidth as the public network.
> 
> The public network is between clients and gateways on the one side to
> talk to monitors, managers, MDS nodes, OSD nodes. It is also used by
> monitors, managers, and MDS nodes to talk with OSD nodes.
> 
> move 2.9 in here.
> 
> Change in 2.3.1
> "back-end" -> "internal"

@dmbyte - over to you on this one :)

lproven:
> If you kept it as 2 seperate NICs for the 2 networks. then the max bandwidth for a network is 1 NIC. And if a NIC or Switch does down, you loose that network on the host.
> 
> On the other hand, if you bond them, then the max bandwidth for a network is 2xNics. Yes you have 2 networks sharing 2 nics, So maybe 2 busy networks could cause issuse because not as good isolation, but still I think the pros out weight the cons. Also we're probably talking about tuning for performance. And this maximises the bandwidth for user data.
> Further, if a NIC or Switch goes down, we only loose some bandwidth, non of the networks are down on the host. So a bit of performance degregation but no loss of the node (which would be much worse).
> Also means you could potentially do things like upgrade network infrastructure while keeping nodes up.

I thought hard about this and yes, that is what I worked out.

However it does not answer my question.

> Sure it may look confusing tho, bond 2 physical nics. Then, from what it looks like in linux interfaces, split the bond into 2 virtual vlan nics. but in essence it's just giving the same bond to 2 networks.

I do not know much about VLANs -- it is a technology I have never used. *How* is the bonded pair assigned to 2 different networks? Even if it is very brief, I think this needs to be clarified.

dmbyte:
I will think on how to try and explain VLAN tagging (which is what creating a VLAN interface does) to people who don’t understand it.  Usually when I have to do so, I have an hour and a whiteboard.


David Byte
Sr. Technology Strategist
SCE Enterprise Linux
SCE Enterprise Storage
Alliances and SUSE Embedded
dbyte@suse.com<mailto:dbyte@suse.com>
918.528.4422

asettle:
Before I can go forward, I know that @dmbyte and @ajaeger have requested two different ways for this information to be presented. 
The current PR is David's suggestion, which is mostly a rewrite of the current introductory paragraph. Andreas would prefer a new section titled "overview". David initially stepped in to help because there was a need for someone to help write the content. 
Could we please focus on coming to a consensus on this, and then moving forward with generating the content for this section. We can step back and discuss VLAN tagging once we've agreed on how we're going to go forward :) 
Thanks,
Alex

xml/deployment_hwrecommend.xml
dmbyte: correct.

xml/deployment_hwrecommend.xml
```diff
@@ -23,6 +23,172 @@
...
+      </para>
+    </listitem>
+  </itemizedlist>
+   <para>
+     The trusted internal network is the back-end network between the OSD
+     nodes for replication and balancing, it needs roughly three times the
+     bandwidth as the public network.
+     The public network is between clients and gateways on the one side to
+     talk to monitors, managers, MDS nodes, OSD nodes. It is also used by
+     monitors, managers, and MDS nodes to talk with OSD nodes.
```
ajaeger: Can we have a graphic showing the networks?

xml/deployment_hwrecommend.xml
```diff
@@ -23,6 +23,172 @@
...
+    <listitem>
+      <para>
+        Client networks for gateways, these are optional.
+      </para>
+    </listitem>
+  </itemizedlist>
+   <para>
+     The trusted internal network is the back-end network between the OSD
+     nodes for replication and balancing, it needs roughly three times the
+     bandwidth as the public network.
```
jan--f: Hmm I need to nitpick this a little. I'd propose to change this to
```
The trusted internal network is the back-end network between the OSD nodes for replication,
re-balancing and recovery. Ideally this network has twice the bandwidth of the public network
since with default 3-way replication the primary OSD sends 2 copies to other OSDs via this network.

File: xml/deployment_hwrecommend.xml [hunk 4]
```diff
@@ -23,6 +23,172 @@
...
+     <imageobject role="html">
+      <imagedata fileref="network-overview-diagram.png" width="70%" format="PNG"/>
+     </imageobject>
+    </mediaobject>
+   </figure>
+
+  <sect2 xml:id="ceph-install-ceph-deploy-network">
+   <title>Network Recommendations</title>
+   <para>
+    For the &ceph; cluster network environment, we recommend two bonded 25 GbE (or faster)
```
jan--f:
```suggestion
    For the &ceph; network environment, we recommend two bonded 25 GbE (or faster)
```
cluster otherwise clashes with cluster network further down.

xml/deployment_hwrecommend.xml
dmbyte: I'd argue that.  From a simplicity perspective, it would be better to put all four ports into a single bond and continue to use VLANs.  This dramatically improves the serviceability by eliminating trying to figure out which physical port is some randomly assigned eth device.

xml/deployment_hwrecommend.xml
```diff
@@ -23,6 +23,172 @@
   basis. If several processes are located on the same machine, the CPU, RAM,
   disk and network requirements need to be added up.
  </para>
+
+<sect1 xml:id="network-overview">
+  <title>Network Overview</title>
+  <para>
+    Ceph has several logical networks:
```
tbazant:
```suggestion
    &ceph; has several logical networks:
```

xml/deployment_hwrecommend.xml
```diff
@@ -23,6 +23,172 @@
...
+   </important>
+
+   <tip>
+    <title>Nodes Configured via DHCP</title>
+    <para>
+     If your storage nodes are configured via DHCP, the default timeouts may not
+     be sufficient for the network to be configured correctly before the various
+     &ceph; daemons start. If this happens, the &ceph; MONs and OSDs will not
+     start correctly (running <command>systemctl status ceph\*</command> will
+     result in "unable to bind" errors) To avoid this issue, we recommend
```
tbazant:
```suggestion
     result in "unable to bind" errors). To avoid this issue, we recommend
```

xml/deployment_hwrecommend.xml
jan--f: Fair enough regarding the VLANs, no objections againt that. However I'm arguing the cluster network does not need to be 3xbandwidth of the public network.

I think this is a change that should be added. I'm ok with recommending bonds and VLANs over physically separate networks.

ajaeger: The text is saying 2xbandwith of public network, not 3 times.

Do you have a better rule of thumb? Or a suggestion for reformulation?

ajaeger: Now I'm getting it - you propose to change "3x" to "2x", correct?

jan--f: correct. One replica wirth of data comes via the public network from the client to the primary. The primary then writes two replicas to other OSDs. All that under the assumption of 3way-replication of course.

ajaeger: When I wrote 3 I was thinking of rebalancing as well. We can write 2 here - just wanted some rule of thumb.

PR #174, opened 2020-02-05
Adding straw to straw2 verification step (BSC#1162242)

asettle:
This also completes JSC#356

### DIFF
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -137,7 +137,7 @@
           Changing a bucket type from <literal>straw</literal> to <literal>straw2</literal> results
           in a small amount of data movement, depending on how much the bucket
           item weights vary from each other. When the weights are all the same, no data
-          will move. When an item weights vary significantly there will be more movement.
+          will move. When an item's weight varies significantly there will be more movement.
           To migrate, execute:
         </para>
 <screen>&prompt.cephuser;ceph osd getcrushmap -o backup-crushmap
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -121,13 +121,13 @@
       <para>
        Run the <command>salt-run upgrade.check</command> command to verify that
        the cluster uses the new bucket type <literal>straw2</literal> and that
-       the &adm; is not a storage node. <literal>straw2</literal>  is the default
+       the &adm; is not a storage node. The default is <literal>straw2</literal>
        for any newly created buckets.
       </para>
       <important>
         <para>
           The new <literal>straw2</literal> bucket type fixes several limitations
-          in the original <literal>straw</literal> bucket. The previous
+          in the original <literal>straw</literal> bucket type. The previous
           <literal>straw</literal> buckets would change some mappings that should
           have changed when a weight was adjusted. <literal>straw2</literal> achieves
           the original goal of only changing mappings to or from the bucket item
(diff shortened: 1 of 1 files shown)


xml/admin_ceph_upgrade.xml
```diff
@@ -121,8 +121,36 @@
...
+          <literal>straw</literal> buckets would change some mappings that should
+          have changed when a weight was adjusted. <literal>straw2</literal> achieves
+          the original goal of only changing mappings to or from the bucket item
+          whose weight has changed.
+        </para>
+        <para>
+          Changing a bucket type from <literal>straw</literal> to <literal>straw2</literal> results
+          in a small amount of data movement, depending on how much the bucket
+          item weights vary from each other. When the weights are all the same, no data
+          will move. When an item weights vary significantly there will be more movement.
```
lproven: "When an item weights vary"?

I would guess either "When item weights vary" or "when an item weight varies"?

xml/admin_ceph_upgrade.xml
```diff
@@ -121,8 +121,36 @@
       <para>
        Run the <command>salt-run upgrade.check</command> command to verify that
        the cluster uses the new bucket type <literal>straw2</literal> and that
-       the &adm; is not a storage node.
+       the &adm; is not a storage node. <literal>straw2</literal>  is the default
```
lproven: Being super-pedantic: 
• 2 spaces after the straw2 literal
• I do not like a sentence beginning with an uncapitalized word. Can we change it to "The default is `straw2` for any newly-created buckets"?

xml/admin_ceph_upgrade.xml
```diff
@@ -121,8 +121,36 @@
...
+       for any newly created buckets.
       </para>
+      <important>
+        <para>
+          The new <literal>straw2</literal> bucket type fixes several limitations
+          in the original <literal>straw</literal> bucket. The previous
```
lproven:
```suggestion
          in the original <literal>straw</literal> bucket type. The previous
```

xml/admin_ceph_upgrade.xml
```diff
@@ -121,8 +121,36 @@
...
+          <literal>straw</literal> buckets would change some mappings that should
+          have changed when a weight was adjusted. <literal>straw2</literal> achieves
```
lproven: See earlier comment but it's not so easily reworded here. :-(

xml/admin_ceph_upgrade.xml
```diff
@@ -121,8 +121,36 @@
...
+        </para>
+<screen>&prompt.cephuser;ceph osd getcrushmap -o backup-crushmap
+&prompt.cephuser;ceph osd crush set-all-straw-buckets-to-straw2</screen>
+        <para>
+          If there are problems, you can revert this change with:
+        </para>
+<screen>&prompt.cephuser;ceph osd setcrushmap -i backup-crushmap</screen>
+        <para>
+          Moving to <literal>straw2</literal> buckets unlocks a few recent features,
+          such as the <literal>crush-compat</literal> balancer mode that was added back in Luminous.
```
lproven: Ambiguous. Does it mean "was restored to the software in $VERSION" or does it mean "this feature was added way back in $VERSION"?

The former I would prefer to clarify with a version number. The latter I feel is too informal and would reword, possibly just by losing the word "back" and again adding a version number.

Both Ubuntu and macOS use codenames for every version, and many people find them confusing. I still have to check most macOS versions every time -- most recently today -- and after ±32 versions of Ubuntu now I have almost totally lost track of the codenames and now avoid them. For some, working with them every day, they are 2nd nature; for others, they are confusing. So I'd like to see version numbers _as well_.

xml/admin_ceph_upgrade.xml
asettle: I'm going to leave this one, sorry.

xml/admin_ceph_upgrade.xml
asettle: Luminous is the "version number" and contextually makes sense for the product and user. I have removed "back" as per your suggestion.

PR #177, opened 2020-02-11
Adding important note for keyring compromise (BSC#1112917)

asettle:


xml/admin_cephx.xml
```diff
@@ -730,6 +730,13 @@ exported keyring for client.admin
     </para>
 <screen>&prompt.cephuser;ceph-authtool /etc/ceph/ceph.keyring \
   --import-keyring /etc/ceph/ceph.client.admin.keyring</screen>
+  <important>
+    <para>
+      If your keyring is compromised, delete your key from the
+      <filename>/etc/ceph</filename> directory and recreate a new key using the same
```
Martin-Weiss: I am not sure if this is enough - I believe we have to mention to delete the key not from /etc/ceph - we have to delete it via "ceph auth del"..

PR #179, opened 2020-02-17
SMB gatways in SLE-HA setup enhancements (bsc#1123896)

tbazant:
This update inserts information how to use RBD images together with SBD as a fencing device.

### DIFF
--- a/xml/deployment_cifs.xml
+++ b/xml/deployment_cifs.xml
@@ -352,7 +352,7 @@
       </step>
       <step>
        <para>
-        Map the RBD image <literal>rbd</literal> on both &sgw;s via
+        Map the RBD image <literal>sbd01</literal> on both &sgw;s via
         <systemitem class="daemon">rbdmap.service</systemitem>.
        </para>
        <para>
--- a/xml/deployment_cifs.xml
+++ b/xml/deployment_cifs.xml
@@ -235,12 +235,14 @@
       for more details.
      </para>
      <para>
-      Create a single RBD image called <literal>rbd</literal> on the &adm; if
-      it does not already exist. Adjust the <option>pg_num</option> number of
-      <literal>100</literal> to your environment:
+      If it does not yet exists, create an RBD pool called
+      <literal>rbd</literal> (see <xref linkend="ceph-pools-operate-add-pool"
+       />) and associate it with <literal>rbd</literal> (see <xref
+       linkend="ceph-pools-associate" />). Then create a related RBD image
+      called <literal>sbd01</literal>:
      </para>
 <screen>
-&prompt.cephuser;ceph osd pool create rbd 100 100 replicated
+&prompt.cephuser;ceph osd pool create rbd <replaceable>PG_NUM</replaceable> <replaceable>PGP_NUM</replaceable> replicated
 &prompt.cephuser;ceph osd pool application enable rbd rbd
 &prompt.cephuser;rbd -p rbd create sbd01 --size 64M --image-shared
 </screen>
--- a/xml/deployment_cifs.xml
+++ b/xml/deployment_cifs.xml
@@ -215,6 +215,7 @@
     <link xlink:href="https://documentation.suse.com/sle-ha/15-SP1/single-html/SLE-HA-install-quick/"/>.
    </para>
    <procedure xml:id="proc-sec-ses-cifs-ha">
+
     <step>
      <para>
       Create a &sgw; specific keyring on the &adm; and copy it to both nodes:
@@ -223,6 +224,25 @@
     osd 'allow *' mds 'allow *' -o ceph.client.samba.gw.keyring
 &prompt.smaster;<command>scp</command> ceph.client.samba.gw.keyring &si.earth;:/etc/ceph/
 &prompt.smaster;<command>scp</command> ceph.client.samba.gw.keyring &si.mars;:/etc/ceph/</screen>
+    </step>
+    <step>
+     <para>
+      SLE-HA setup requires a fencing device for split brain avoidance.  For
+      this purpose, you can use a &ceph; RBD image with Stonith Block Device
+      (SBD). Refer to <link
+       xlink:href="https://documentation.suse.com/sle-ha/15-SP1/single-html/SLE-HA-guide/#sec-ha-storage-protect-fencing-setup"/>
+      for more details
+     </para>
+     <para>
+      Create a single RBD image called <literal>rbd</literal> on the
+      &adm; while adjusting the <option>pg_num</option> number of
+      <literal>100</literal> to your environment:
+     </para>
+<screen>
+&prompt.cephuser;ceph osd pool create rbd 100 100 replicated
+&prompt.cephuser;ceph osd pool application enable rbd rbd
+&prompt.cephuser;rbd -p rbd create sbd01 --size 64M --image-shared
+</screen>
     </step>
     <step>
      <para>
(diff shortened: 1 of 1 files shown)


xml/deployment_cifs.xml
```diff
@@ -215,14 +215,34 @@
...
 <screen>&prompt.cephuser;<command>ceph</command> auth get-or-create client.samba.gw mon 'allow r' \
     osd 'allow *' mds 'allow *' -o ceph.client.samba.gw.keyring
-&prompt.smaster;<command>scp</command> ceph.client.samba.gw.keyring &si.earth;:/etc/ceph/
-&prompt.smaster;<command>scp</command> ceph.client.samba.gw.keyring &si.mars;:/etc/ceph/</screen>
+&prompt.cephuser;<command>scp</command> ceph.client.samba.gw.keyring &si.earth;:/etc/ceph/
+&prompt.cephuser;<command>scp</command> ceph.client.samba.gw.keyring &si.mars;:/etc/ceph/</screen>
+    </step>
+    <step>
+     <para>
+      SLE-HA setup requires a fencing device for split brain avoidance.  For
```
asettle:
```suggestion
      SLE-HA setup requires a fencing device for split brain avoidance. For

Comment (asettle) [reply xml/deployment_cifs.xml]: I have no idea what "split brain avoidance" means?!

File: xml/deployment_cifs.xml [hunk 2]
```diff
@@ -215,14 +215,34 @@
...
+      this purpose, you can use a &ceph; RBD image with Stonith Block Device
+      (SBD). Refer to <link
+       xlink:href="https://documentation.suse.com/sle-ha/15-SP1/single-html/SLE-HA-guide/#sec-ha-storage-protect-fencing-setup"/>
+      for more details
```
asettle:
```suggestion
      for more details.

File: xml/deployment_cifs.xml [hunk 3]
```diff
@@ -215,14 +215,34 @@
...
+     </para>
+     <para>
+      Create a single RBD image called <literal>rbd</literal> on the
+      &adm; while adjusting the <option>pg_num</option> number of
```
asettle:
```suggestion
      &adm; and adjust the <option>pg_num</option> number of

File: xml/deployment_cifs.xml [hunk 4]
```diff
@@ -328,19 +349,33 @@
...
+</screen>
+      <para>
+       The <filename>/dev/rbd/rbd/sbd01</filename> device should be available
+       on both &sgw;s.
+      </para>
       </step>
       <step>
        <para>
-        Let &si.mars; join the cluster:
+        Initialize the cluster on &si.earth; and let &si.mars; join it. When
```
asettle: Maybe say "A request to allow &si.mars; join the cluster will appear, when asked..." otherwise it's a bit confusing

xml/deployment_cifs.xml
ddiss: I think it makes sense to refer to the SLE-HA documentation, where possible. For "split brain" they have a link to an external site:
https://techthoughts.typepad.com/managing_computers/2007/10/split-brain-quo.html

xml/deployment_cifs.xml
```diff
@@ -258,14 +278,14 @@
...
+      On &si.earth;, create the configuration files for Samba. They will later
+      be automatically synchronized to &si.mars;.
      </para>
      <substeps>
       <step>
        <para>
-        In <filename>/etc/ctdb/nodes</filename> insert all nodes that contain
-        all private IP addresses of each node in the cluster:
+        Insert all nodes that contain all private IP addresses of each node in
+        the cluster in the <filename>/etc/ctdb/nodes</filename> file:
```
ddiss: I find this a little difficult to understand, in both current and proposed forms. I'll try to think of something else here and submit separately.

tbazant: ok, leaving it up to you then

xml/deployment_cifs.xml
```diff
@@ -328,19 +352,40 @@
       </step>
       <step>
        <para>
-        Initialize the cluster on &si.earth;:
+        Map the RBD image <literal>rbd</literal> on both &sgw;s via
```
ddiss: Sorry, I missed this previously - the RBD image name here should also be _sbd01_ . Looks good otherwise.

xml/deployment_cifs.xml
ddiss: The upstream ctdb man page (https://ctdb.samba.org/manpages/ctdb.7.html ) does a good job of describing the purpose of the CTDB "private network" and the contents of the _nodes_ file:
```
Private address

Each node is configured with a unique, permanently assigned private address. This address is configured by the operating system. This address uniquely identifies a physical node in the cluster and is the address that CTDB daemons will use to communicate with the CTDB daemons on other nodes.

Private addresses are listed in the file /usr/local/etc/ctdb/nodes). This file contains the list of private addresses for all nodes in the cluster, one per line. This file must be the same on all nodes in the cluster.

Some users like to put this configuration file in their cluster filesystem. A symbolic link should be used in this case.

Private addresses should not be used by clients to connect to services provided by the cluster.

It is strongly recommended that the private addresses are configured on a private network that is separate from client networks. This is because the CTDB protocol is both unauthenticated and unencrypted. If clients share the private network then steps need to be taken to stop injection of packets to relevant ports on the private addresses. It is also likely that CTDB protocol traffic between nodes could leak sensitive information if it can be intercepted.

Example /usr/local/etc/ctdb/nodes for a four node cluster:

192.168.1.1
192.168.1.2
192.168.1.3
192.168.1.4
```

I propose that we use this verbatim, with "node" change to "Samba Gateway node". Can you insert that with this series, or should I submit a follow-up?

xml/deployment_cifs.xml
```diff
@@ -215,14 +215,37 @@
...
+     <para>
+      SLE-HA setup requires a fencing device to avoid a <emphasis>split
+       brain</emphasis> situation when active cluster nodes become
+      unsynchronized. For this purpose, you can use a &ceph; RBD image with
+      Stonith Block Device (SBD). Refer to <link
+       xlink:href="https://documentation.suse.com/sle-ha/15-SP1/single-html/SLE-HA-guide/#sec-ha-storage-protect-fencing-setup"/>
+      for more details.
+     </para>
+     <para>
+      If it does not yet exists, create an RBD pool called
```
asettle:
```suggestion
      If it does not yet exist, create an RBD pool called

File: xml/deployment_cifs.xml [hunk 8]
```diff
@@ -258,14 +283,15 @@
     </step>
     <step>
      <para>
-      On &si.earth; create the configuration files for Samba. They will later
-      automatically be synchronized to &si.mars;.
+      On &si.earth;, create the configuration files for Samba. They will later
+      be automatically synchronized to &si.mars;.
```
asettle:
```suggestion
      automatically synchronize to &si.mars;.

File: xml/deployment_cifs.xml [hunk 9]
```diff
@@ -328,19 +355,40 @@
...
-&prompt.mars;<command>ha-cluster-join</command> -c earth
-      </screen>
+        Initialize the cluster on &si.earth; and let &si.mars; join it.
+               </para>
+<screen>&prompt.earth;<command>ha-cluster-init</command></screen>
+<screen>&prompt.mars;<command>ha-cluster-join</command> -c earth</screen>
+       <important>
+        <para>
+         During the process of initialization and joining the cluster, you will
+         be interactively asked whether to use SBD. Confirm with 'y' and then
```
asettle:
```suggestion
         be interactively asked whether to use SBD. Confirm with <option>y</option> and then

PR #188, opened 2020-02-25
Update authors.xml

asettle:


asettle:
@fsundermeyer - as Sven is no longer part of the company, do we remove him from this file? Or do we keep him as he was a contributor? If we keep him, we need to have a line for when we remove authors after they move on from the company/team.

PR #190, opened 2020-03-02
Updated Ops & Admin Guide (jsc#ses 1397)

tbazant:
For SES7, the topics in the administration guide will be re-sorted, and some of them will be moved to different guide as they better match the topics. This update finalizes the structure + creates a skeleton for a new Tuning Guide with 2 included chapters to fullfill references.


tbazant:
Locally, this branch validates OK, not sure why CI fails on it :-(

asettle:
> Locally, this branch validates OK, not sure why CI fails on it :-(

It doesn't validate locally for me. It's failing with the same error as the CI build. The concern is the cross-ref with `storage-tips-ceph-btrfs-subvol`. Without going too indepth, I think your issue is that you've removed `admin_ceph_tips.xml` from `book_storage_admin.xml` TOC, but you haven't actually deleted the file. So where it is referenced, it's still searching for the link and failing.

PR #197, opened 2020-03-05
Purging is no longer required, remove 4th step (BSC#1165658)

asettle:


xml/admin_ceph_gateway.xml
```diff
@@ -2045,17 +2045,6 @@ GET /<replaceable>BUCKET</replaceable>?mdsearch
        </para>
```
theanalyst: Drop the last sentence here, that was needed only to reference the now deleted sentence

theanalyst: Also the title should be just "resharding the bucket index" drop pool

xml/admin_ceph_gateway.xml
```diff
@@ -2040,22 +2040,10 @@ GET /<replaceable>BUCKET</replaceable>?mdsearch
        <title>Old Bucket ID</title>
```
theanalyst: the command in L2035 `radosgw-admin bucket reshard...` (the bucket is missing)

PR #199, opened 2020-03-06
Clarifying CPU device usage (BSC#1159724)

asettle:


xml/deployment_hwrecommend.xml
```diff
@@ -252,7 +252,8 @@
    <itemizedlist mark="bullet" spacing="normal">
     <listitem>
      <para>
-      CPU recommendations:
+      The follow CPU recommendations accounts for all devices
```
jan--f:
```suggestion
      The following CPU recommendations accounts for all devices
```

xml/deployment_hwrecommend.xml
```diff
@@ -252,7 +252,8 @@
    <itemizedlist mark="bullet" spacing="normal">
     <listitem>
      <para>
-      CPU recommendations:
+      The following CPU recommendations accounts for all devices
```
tbazant:
```suggestion
      The following CPU recommendations account for all devices
```

ajaeger: Isn't it more that it's independent of usage of device?
So, "The following CPU recommendations account for devices independent of usage"?

PR #200, opened 2020-03-09
Updating dashboard node config (BSC#1166052)

asettle:


xml/admin_about.xml
```diff
@@ -266,7 +263,10 @@
        The &ceph; manager (MGR) collects the state information from the whole
```
lproven: I think there is an entity for this? &mgr;

tbazant: +1

xml/admin_about.xml
```diff
@@ -266,7 +263,10 @@
...
        cluster. The &ceph; manager daemon runs alongside the monitor daemons.
        It provides additional monitoring, and interfaces the external
-       monitoring and management systems.
+       monitoring and management systems. It includes other services
+       as well, for example the &dashboard; Web UI with the
+       <emphasis>&grafana;</emphasis> dashboard backed by the
```
LenzGr: Actually, both Prometheus and Grafana are independent instances that do not have any direct relation to the Dashboard or the Ceph Mgr daemon. The Dashboard is capable of "consuming"/embedding Grafana dashboards, which in turn uses Prometheus to collect the metrics that it visualizes.

asettle: I think it's best to just remove everything after "web UI" in that case :) thanks Lenz

PR #215, opened 2020-03-26
Adding chown instruction for more user friendly exp (BSC#1166417)

asettle:


xml/admin_install_salt.xml
```diff
@@ -726,8 +726,9 @@ done</screen>
      <filename>/srv/pillar/ceph/stack/ceph/cluster.yml</filename> manually and
      set the <option>cluster_network:</option> and
      <option>public_network:</option> options. Note that the file will not be
-     overwritten after you install <package>deepsea</package>.
+     overwritten after you install <package>deepsea</package>. Then, run:
     </para>
+<screen>chown -R salt:salt /srv/pill/ceph/stack</screen>
```
ddiss: I think that should be pillar, not pill, i.e. `/srv/pillar/ceph/stack/`. @swiftgist shouldn't the `/srv/pillar/ceph` directory also be salt-owned?

asettle: That's what I get for copy/paste and no checking. Thanks @ddiss

swiftgist: yeah, did I type that in the original bug? :man_facepalming:

PR #221, opened 2020-04-02
Updating min reqs for containerized deploy (BSC#1165346, BSC# 1165345)

asettle:


xml/containerized-kubernetes.xml
```diff
@@ -87,14 +103,14 @@
   <para>
    The &rook; orchestrator uses configuration files in YAML format called
    <emphasis>manifests</emphasis>. The manifests you need are included in the
-   <package>rook-k8s-yaml</package> RPM package. Install it by running
+   <package>rook-k8s-yaml</package> RPM package. You can find this package
```
denisok, line 106: nope. rook-k8s-yaml package is in SES6 repo. /usr/share/k8s-yaml is the path where you would see Rook manifest after installation of that package.

asettle: @denisok what's the path? (is there a path?)

denisok: so, section 5.3 of deployment guide says:
...
 > The following procedure describes the cluster preparation in detail. 
...
> Install and register SUSE Linux Enterprise Server 15 SP1 together with the SUSE Enterprise Storage 6 extension on each node of the cluster. 

In case of CaaSP, we don't need SLE15 SP! as we are saying:
>  A running SUSE CaaS Platform 4 cluster. 
which already implies SLE15SP1 and CaaSP repos.

But what we still need is "SUSE Enterprise Storage 6 extension"
That should be configured on kubernetes management node.
...

So we need to say that on management node these 2 repos needs to be configured too install rook-k8s-yaml package:
SUSE-Enterprise-Storage-6-Pool
 SUSE-Enterprise-Storage-6-Updates

How to configure those repos, I haven't find that in a Deployment or other guide. That seems to be more high level SUSE document. It could be done via SCC or by adding repositories.
6.8.1 has some definition but installing from DVD, which is not applicable here.

xml/containerized-kubernetes.xml
```diff
@@ -62,23 +60,41 @@
...
-   Before you start deploying, you need to have:
+   The minimum requirements and prerequisites to deploy &productname; &productnumber;
+   on top of &casp; 4 &kube; cluster are as follows:
   </para>
-
   <itemizedlist>
    <listitem>
     <para>
-     A running &casp; 4 cluster.
+     A running &casp; 4 cluster. You need to have an SCC account with an
```
BlaineEXE, line 70: Would suggest some wording similar to "... account with a SES6 subscription. You can activate a 60-day..." because non-eval subscriptions also work.

xml/containerized-kubernetes.xml
```diff
@@ -62,23 +60,41 @@
...
     <para>
-     &casp; 4 worker nodes with a number of extra disks attached as storage for
-     the &ceph; cluster.
+     At least 3 &casp; worker nodes with a number of extra disks attached as
+     storage for the &ceph; cluster. We recommend 4 &casp; worker nodes.
     </para>
    </listitem>
+   <listitem>
+     <para>
+       At least 1 OSD per worker node, with a minimum disk size of 5GB.
```
lproven, line 83:
```suggestion
       At least one OSD per worker node, with a minimum disk size of 5&nbsp;GB.
```

xml/containerized-kubernetes.xml
asettle: "You can activate a 60-day free Evaluation here <link> because non-evaluation subscriptions also work."

Uhh, is that what you mean to say  @BlaineEXE ?

xml/containerized-kubernetes.xml
```diff
@@ -95,6 +95,13 @@
        &casp; cluster. We recommend using the &casp; master node as the workstation.
      </para>
    </listitem>
+   <listitem>
+      <para>
+        Ensure that the <literal>SUSE-Enterprise-Storage-6-Pool</literal> and
+        <literal>SUSE-Enterprise-Storage-6-Updates</literal> repos are configured on
```
lproven, line 101: I would prefer the slightly more formal "repositories" over "repos"...

xml/containerized-kubernetes.xml
```diff
@@ -87,14 +110,14 @@
   <para>
    The &rook; orchestrator uses configuration files in YAML format called
    <emphasis>manifests</emphasis>. The manifests you need are included in the
-   <package>rook-k8s-yaml</package> RPM package. Install it by running
+   <package>rook-k8s-yaml</package> RPM package. You can find this package
```
denisok, line 113: you forgot to remove "You can find this package in <filename>/usr/share/k8s-yaml" it should be " That package is in SES6 repos"

xml/containerized-kubernetes.xml
```diff
@@ -62,23 +60,48 @@
...
-   Before you start deploying, you need to have:
+   The minimum requirements and prerequisites to deploy &productname; &productnumber;
+   on top of &casp; 4 &kube; cluster are as follows:
   </para>
-
   <itemizedlist>
    <listitem>
     <para>
-     A running &casp; 4 cluster.
+     A running &casp; 4 cluster. You need to have an account with an
```
tbazant, line 70:
```suggestion
     A running &casp; 4 cluster. You need to have an account with a
```

xml/containerized-kubernetes.xml
```diff
@@ -62,23 +60,48 @@
...
+     &productname; &productnumber; subscription. You can activate a 60-day free evaluation
```
tbazant, line 71: Should this be rather CaaSP?
```suggestion
     &casp; subscription. You can activate a 60-day free evaluation
```

xml/containerized-kubernetes.xml
```diff
@@ -62,23 +60,48 @@
...
+   <listitem>
+     <para>
+       Access to a workstation that has access via <literal>kubectl</literal> to the
+       &casp; cluster. We recommend using the &casp; master node as the workstation.
+     </para>
+   </listitem>
+   <listitem>
+      <para>
+        Ensure that the <literal>SUSE-Enterprise-Storage-6-Pool</literal> and
+        <literal>SUSE-Enterprise-Storage-6-Updates</literal> repos are configured on
```
tbazant, line 101: +1
```suggestion
        <literal>SUSE-Enterprise-Storage-6-Updates</literal> repositories are configured on
```

xml/containerized-kubernetes.xml
```diff
@@ -62,23 +60,48 @@
...
+     A running &casp; 4 cluster. You need to have an account with a
+     &casp; subscription. You can activate a 60-day free evaluation
+     here <link xlink:href="https://www.suse.com/products/caas-platform/download/MkpwEt3Ub98~/?campaign_name=Eval:_CaaSP_4"/>.
     </para>
    </listitem>
    <listitem>
     <para>
-     &casp; 4 worker nodes with a number of extra disks attached as storage for
-     the &ceph; cluster.
+     At least 3 &casp; worker nodes with a number of extra disks attached as
```
lproven, line 77: I suggest:

> Three &casp; worker nodes
&
> We recommend four

Numbers under 10 written out in full is the most common style.

I think "a number of extra disks" is too vague.

How about:

... worker nodes, with at least one additional disk attached to each node as storage...

xml/containerized-kubernetes.xml
```diff
@@ -62,23 +60,48 @@
...
+   </listitem>
+   <listitem>
+     <para>
+       Access to &productname; &productnumber;. You can get a trial
+       subscription from here <link xlink:href="https://www.suse.com/products/suse-enterprise-storage/download/"/>.
+     </para>
+   </listitem>
+   <listitem>
+     <para>
+       Access to a workstation that has access via <literal>kubectl</literal> to the
```
lproven, line 94: Suggestion:
"Access to a workstation that has access to the cluster via kubectl" -- that word order makes more sense to me. What do you think?

The master node as the workstation? Really? It can orchestrate itself? I did not know that was possible. Does it not also mean the master must run X.11 and so on? That does not sound like a good idea, but I defer to the experts.

asettle: @denisok ?

denisok: I don't think we should be picky on this, especially for tech preview. fine I am.

PR #222, opened 2020-04-02
minor Samba gw admin fixes

ddiss:


asettle:
@ddiss it was having a barney because you had the audacity to include a <screen> tag in a <para> ;) Anyway, I pushed an update to fix that and approved. Thanks!

ddiss:
> @ddiss it was having a barney because you had the audacity to include a tag in a ;) Anyway, I pushed an update to fix that and approved. Thanks!

Oops, sorry. Thanks for the follow up fix!

asettle:
> > @ddiss it was having a barney because you had the audacity to include a tag in a ;) Anyway, I pushed an update to fix that and approved. Thanks!
> 
> Oops, sorry. Thanks for the follow up fix!

NP! Merge and backport at will :) no need to create a PR for the backports. Just push directly to the branch.

Issue #223, opened 2020-04-02
[doc] 20.1.2  Samba Gateway Example missing kernel backed share instructions


ddiss:
20.1.2  Single Gateway Example

http://docserv.nue.suse.com/documents/SES_6/ses-admin/single-html/#sec-ses-cifs-example

We currently only provide instructions for using Samba with the ceph VFS backend. Instructions for using Samba atop a kernel CephFS mount should also be present.

PR #224, opened 2020-04-02
deployment_smb: kernel backed share instructions

ddiss:
Fixes: https://github.com/SUSE/doc-ses/issues/223
Signed-off-by: David Disseldorp <ddiss@suse.de>

### DIFF
--- a/xml/deployment_cifs.xml
+++ b/xml/deployment_cifs.xml
@@ -127,6 +127,25 @@
   oplocks = no
   kernel share modes = no
  </screen>
+     <para>
+      The above share configuration uses the <systemitem>vfs_ceph</systemitem>
+      module for &samba; to communicate with the &ceph; cluster. As a
+      higher-performing but less-flexible alternative, &samba; can instead be
+      configured to use the Linux kernel &cephfs; client, as shown below:
+     </para>
+<screen>
+[<replaceable>SHARE_NAME</replaceable>]
+  path = <replaceable>CEPHFS_MOUNT</replaceable>
+  # "vfs objects = ceph" and "ceph: config_file / user_id" are dropped
+  read only = no
+  oplocks = no
+  kernel share modes = no
+</screen>
+     <para>
+      The <replaceable>CEPHFS_MOUNT</replaceable> path above must be mounted
+      prior to starting &samba; with a kernel &cephfs; share configuration. See
+      <xref linkend="ceph-cephfs-cephfs-fstab"/>.
+     </para>
      <tip>
       <title>Oplocks and Share Modes</title>
       <para>
@@ -156,13 +175,10 @@
 &prompt.root;smbpasswd -a <replaceable>USERNAME</replaceable>
 </screen>
       <para>
-       Since <systemitem>vfs_ceph</systemitem> does not require a file system
-       mount, the share path is interpreted as an absolute path within the
-       &ceph; file system on the attached &ceph; cluster. For successful share
-       I/O, the path's access control list (ACL) needs to permit access from
-       the mapped user for the given Samba client. You can modify the ACL by
-       temporarily mounting via the &cephfs; kernel client and using the
-       <command>chmod</command>, <command>chown</command>, or
+       For successful I/O, the share path's access control list (ACL) needs to
+       permit access to the user connected via &samba;. You can
+       modify the ACL by temporarily mounting via the &cephfs; kernel client
+       and using the <command>chmod</command>, <command>chown</command>, or
        <command>setfacl</command> utilities against the share path. For
        example, to permit access for all users, run:
       </para>


xml/deployment_cifs.xml
```diff
@@ -127,6 +127,25 @@
   oplocks = no
   kernel share modes = no
  </screen>
+     <para>
+      The above share configuration uses the <systemitem>vfs_ceph</systemitem>
+      module for &samba; to communicate with the Ceph cluster. As a more
```
asettle, line 132:
```suggestion
      module for &samba; to communicate with the& eph; cluster. As a more

File: xml/deployment_cifs.xml [hunk 2]
```diff
@@ -156,13 +175,10 @@
...
       <para>
-       Since <systemitem>vfs_ceph</systemitem> does not require a file system
-       mount, the share path is interpreted as an absolute path within the
-       &ceph; file system on the attached &ceph; cluster. For successful share
-       I/O, the path's access control list (ACL) needs to permit access from
-       the mapped user for the given Samba client. You can modify the ACL by
-       temporarily mounting via the &cephfs; kernel client and using the
-       <command>chmod</command>, <command>chown</command>, or
+       For successful I/O, the share path's access control list (ACL) needs to
+       permit access from the mapped user for the given Samba client. You can
```
asettle, line 179:
```suggestion
permit access from the mapped user for the given &samba; client. You can

**Comment by ddiss** [comment]
@asettle thanks for the feedback. I've pushed an update. Is there a service that can render the documentation from this source branch, or do I need to build it locally?

File: xml/deployment_cifs.xml [hunk 3]
```diff
@@ -127,6 +127,25 @@
   oplocks = no
   kernel share modes = no
  </screen>
+     <para>
+      The above share configuration uses the <systemitem>vfs_ceph</systemitem>
+      module for &samba; to communicate with the &ceph; cluster. As a more
```
lproven, line 132: I suggest

> as a higher-performance but less-flexible alternative

asettle:
> @asettle thanks for the feedback. I've pushed an update. Is there a service that can render the documentation from this source branch, or do I need to build it locally?

Locally I'm afraid. You can use daps2docker or daps locally. It's pretty easy once you're up and running. 

Otherwise, this LGTM @lproven want to take a gander?

xml/deployment_cifs.xml
```diff
@@ -156,13 +175,10 @@
...
       <para>
-       Since <systemitem>vfs_ceph</systemitem> does not require a file system
-       mount, the share path is interpreted as an absolute path within the
-       &ceph; file system on the attached &ceph; cluster. For successful share
-       I/O, the path's access control list (ACL) needs to permit access from
-       the mapped user for the given Samba client. You can modify the ACL by
-       temporarily mounting via the &cephfs; kernel client and using the
-       <command>chmod</command>, <command>chown</command>, or
+       For successful I/O, the share path's access control list (ACL) needs to
+       permit access from the mapped user for the given &samba; client. You can
```
lproven, line 179: This might be my unfamiliarity with Ceph, but I do not understand the phrase

> the mapped user for the given Samba client

How can a user be mapped? I would think that a user is the entity **doing** the mapping to a share, not the entity **being mapped**. I may be wrong.

If I am right, I would suggest:

> permit access by the user mapping to the given Samba client

xml/deployment_cifs.xml
ddiss: My grammar isn't great, but shouldn't that be `a higher performing but less flexible alternative`?

xml/deployment_cifs.xml
ddiss: Good point, use of `map` here is a little confusing. It's a bit of an overloaded term, as it can refer to mapping of a network drive on Windows, etc.
The mapping behaviour that I'm trying to describe is performed by Samba, where it translates between SMB user/group identities (name+SID) and UNIX user/group identities.
I'll have a think about how this can be reworded to be a little clearer. Thanks for your feedback!


, as a reader may think it relates to mapping a

ddiss:
@lproven I've made some changes following your feedback. Please take a look.

xml/deployment_cifs.xml
lproven: That's fine too. In all these instances, the _phrase_ is the adjective: 
* higher-performing
* higher-performance
* less-flexible

Thus the hyphens. Otherwise we would be talking about an alternative that is higher and is also performing, or which is less and is also flexible... which does not make sense.

Compare: three-year-old deer (1 deer, 3YO); three year-old deer (3 deer, all 1YO) three year old deer (does not make sense, as can be seen considering the adjectives separately: what does "3 year deer" mean?)

ddiss: I've pushed a new version with the hyphens added. Thanks

PR #226, opened 2020-04-03
Adding removal of a failed node example (BSC#1167252)

asettle:


asettle:
@swiftgist this is more-or-less a direct translation from the bug report. I feel a little iffy about the explanation, I'm sure it could be better :thinking:

xml/admin_saltcluster.xml
```diff
@@ -351,6 +351,34 @@ role-rgw/cluster/rgw1*.sls</screen>
 &prompt.smaster;salt-run rescinded.ids
 &prompt.smaster;salt-run state.orch ceph.stage.5</screen>
   </example>
+  <example xml:id="ex-failed-node">
+   <title>Removing of a Failed Node</title>
```
lproven, line 355:
```suggestion
   <title>Removal of a Failed Node</title>
```

xml/admin_saltcluster.xml
```diff
@@ -351,6 +351,34 @@ role-rgw/cluster/rgw1*.sls</screen>
...
+   <para>
+     If the &sminion; is not responding and the administrator is unable to
+     resolve the issue, we recommend removing the &salt; key and leaving the
+     server intact and still operating in the capacity as a storage node.
+   </para>
+   <para>When a servers fails (either network, power, or another issue), it means
```
lproven, line 361:
```suggestion
   <para>When a server fails (due to network, power, or other issues), it means
```

swiftgist:
I agree that the paragraph is confusing.  There's three separate points in there:

Failed node
Failed OSD node
Failed OSD node that is still operational

The first applies to any server in a Ceph cluster that fails.  Remove the salt key if the server cannot be brought back online.

The OSD commands are specifically for a storage node that can be run assuming that the server failed rather permanently.  

The third option is one of those "theoretical" options that I cannot imagine happening, but technically  could.  Salt commands are not working, but the server is fine otherwise.

Honestly, to keep this simple, I think I would only address the first two.  Leave the third as a support call because we haven't seen it.  Maybe two headers would make this clearer? Failed node and Failed Storage node?

Does this help?

asettle:
> I agree that the paragraph is confusing. There's three separate points in there:
> 
> Failed node
> Failed OSD node
> Failed OSD node that is still operational
> 
> The first applies to any server in a Ceph cluster that fails. Remove the salt key if the server cannot be brought back online.
> 
> The OSD commands are specifically for a storage node that can be run assuming that the server failed rather permanently.
> 
> The third option is one of those "theoretical" options that I cannot imagine happening, but technically could. Salt commands are not working, but the server is fine otherwise.
> 
> Honestly, to keep this simple, I think I would only address the first two. Leave the third as a support call because we haven't seen it. Maybe two headers would make this clearer? Failed node and Failed Storage node?
> 
> Does this help?

Sort of. So, we're going to split it in half:

1. Removal of a Failed Node
- &prompt.smaster;salt-key -d <replaceable>MINION_ID</replaceable>

2. Removal of a  Failed Storage Node
- &prompt.cephuser;ceph osd purge-actual $id --yes-i-really-mean-it
  &prompt.cephuser;ceph auth del osd.$id

Correct?

swiftgist:
> 
> Sort of. So, we're going to split it in half:
> 
> 1. Removal of a Failed Node
> 
> * &prompt.smaster;salt-key -d MINION_ID
> 
> 1. Removal of a  Failed Storage Node
> 
> * &prompt.cephuser;ceph osd purge-actual $id --yes-i-really-mean-it
>   &prompt.cephuser;ceph auth del osd.$id
> 
> Correct?

Right, but the commands to purge/del the OSDs is run for _each_ OSD on that removed storage node.  So, the loop is still correct.

xml/admin_saltcluster.xml
```diff
@@ -351,6 +351,37 @@ role-rgw/cluster/rgw1*.sls</screen>
 &prompt.smaster;salt-run rescinded.ids
 &prompt.smaster;salt-run state.orch ceph.stage.5</screen>
   </example>
+  <example xml:id="ex-failed-node">
+   <title>Removal of a Failed Node</title>
+   <para>
+     If the &sminion; is not responding and the administrator is unable to
+     resolve the issue, we recommend removing the &salt; key and leaving the
+     server intact and still operating in the capacity as a storage node.
```
swiftgist, line 359: I think the "and leaving... node" can be removed.  The succinct "we recommend removing the salt key" gets to the point without creating other questions for the admin (e.g. what if it's not a storage node?).

asettle:
Alright @swiftgist it's all down to you :D

PR #231, opened 2020-04-07
Adding section on running deepsea to multisite (BSC#1135872)

asettle:


swiftgist:
I will tentatively wait on Martin's example.... since I do not know _how_ he was changing the ceph.conf.

PR #243, opened 2020-04-16
Correcting service names - Closes #241

### LINKED ISSUE #241: [doc] 18.2.9  Starting the Services
18.2.9  Starting the Services

https://documentation.suse.com/ses/6/html/ses-all/cha-ses-cifs.html#cephfs-ad-services

As of SLES15SP1, the names are 'smb.service' , 'nmb.service' and 'winbind.service'.

asettle:


asettle:
@tbazant this should probably also be updated for the SES 7 deployment guide

xml/deployment_cifs.xml
```diff
@@ -953,9 +953,9 @@ group:  files winbind
...
     functioning Unix domain member:
-    <systemitem class="daemon">smbd</systemitem>,
-    <systemitem class="daemon">nmbd</systemitem>, and
-    <systemitem class="daemon">winbindd</systemitem>. Enable and start them by
+    <systemitem class="daemon">smb.service</systemitem>,
+    <systemitem class="daemon">nmb.service</systemitem>, and
+    <systemitem class="daemon">winbind.service</systemitem>. Enable and start them by
     using the following commands:
    </para>
 <screen>
```
ddiss, line 961: As described in the bug, the items above aren't strictly incorrect, as they correspond to the name of the underlying process / daemon. Its the _systemctl_ instructions below which need to be fixed, as the systemd service names don't carry the 'd' suffix.
While we're here, the sudo instructions also appear wrong, as it's used for `systemctl enable` but missing for `systemctl start`.  I'd suggest:
```
&prompt.cephuser.smb;sudo systemctl enable smb.service
&prompt.cephuser.smb;sudo systemctl start smb.service
&prompt.cephuser.smb;sudo systemctl enable nmb.service
&prompt.cephuser.smb;sudo systemctl start nmb.service
&prompt.cephuser.smb;sudo systemctl enable winbind.service
&prompt.cephuser.smb;sudo systemctl start winbind.service
```
Alternatively, perhaps we could use `&prompt.root` and drop sudo altogether?

asettle: I see! Thanks for clarifying :) I'll update now

asettle: I wonder if there was a reason we used 'sudo' rather than root in this case? Would there be any indication of that?

ddiss: Actually, looking closer, I think we also need to provide HA specific instructions here too. The standalone (18.1.2 step 5) and HA (18.1.3 steps 7 & 9) guides already run through enabling and starting the smb and nmb services, so perhaps we can assume that they're up while winbind is being configured?

ddiss: As discussed, I'll raise a separate ticket to track this.

asettle:
> LGTM, thanks. Minor nit: the first patch of the series should be squashed with the last.

We squash and merge :)

PR #253, opened 2020-04-22
Add missing ceph-salt '/time_server/subnet' node

ricardoasmarques:
`/time_server/subnet` exists since https://github.com/ceph/ceph-salt/pull/165

Signed-off-by: Ricardo Marques <rimarques@suse.com>


asettle:
@ricardoasmarques does this need to be backported to SES 6?

PR #282, opened 2020-04-27
Add labeling 'mon' tip (bsc#1170458)

tbazant:
This update adds steps to lable MON nodes for simplified listing and deployment.


tbazant:
Adding rendered output 
[deploy-cephadm-day2_color_en.pdf](https://github.com/SUSE/doc-ses/files/4539834/deploy-cephadm-day2_color_en.pdf)

swiftgist:
I'll double check that `ceph orch apply mon label:mon` is actually needed.  The other part is the entire section could be repeated for mgr.  However, it doesn't work for mds.  So, I'm not sure what to suggest there.

xml/deploy_cephadm.xml
```diff
@@ -835,6 +835,33 @@ o- minions ................................................. [Minions: 5]
 <para>
  -->
 <screen>&prompt.cephuser;ceph orch apply mon ses-min1 ses-min2 ses-min3</screen>
+      <tip>
+       <title>Labels for Nodes of a Specific Service</title>
+       <para>
+        You can mark the nodes that will be used as MONs with the 'mon' label,
```
asettle, line 841:
```suggestion
        You can mark the nodes that will be used as MONs with the <option>mon</option> label,

PR #285, opened 2020-04-29
Updating iperf network bonding explanation - Closes #261

### LINKED ISSUE #261: [doc] 4.1.1  Network Performance - explain a bit about the trunk
4.1.1  Network Performance

https://documentation.suse.com/ses/6/single-html/ses-tuning/#id-1.5.3.3.4.5
---------------------

Depending on the hash mode and client vs. server bandwidth we might have to use multiple iperf in parallel to measure the full combined bandwidth possible. (single connection vs. multiple parallel connections)

asettle:


xml/tuning-os-tuning.xml
```diff
@@ -86,6 +86,20 @@
...
+       address (layer 2+3). Testing with a single session and using only two IP addresses
+       or two MAC addresses with the test will not allow to see and measure the combined
+       bandwidth of multiple NICs in one LACP bond. To test this, you need to run multiple
+       <command>iperf3 -s -p 5901</command> <command>iperf3 -s -p 5902</command> commands
+       on the target host and then execute multiple <command>iperf3</command> on other
+       servers in parallel. For example, <command>iperf3 -c -p 5901</command> on client
+       <literal>server1</literal> and <command>iperf3 -c -p 5902</command> on client <literal>server2</literal>.
+       For more information on network interface bonding, see
+       <link xlink:href="https://documentation.suse.com/sles/15-SP1/html/SLES-all/cha-network.html#sec-network-iface-bonding"/>.
+     </para>
```
lproven, line 102:
```suggestion
     <para>
       We recommend that you configure bonding with <literal>xmit_hash_policy</literal>
       as layer 2 or layer 2+3. Layer 2 bonding distributes the load among the
       interfaces based on source and target MAC address. Layer 2+3 bonding
       distributes the load based on source and target IP addresses.
     </para>
     <para>       
       Testing with a single session and using only two IP addresses, or two MAC
       addresses, will not allow you to see or measure the combined bandwidth
       of multiple NICs in a single LACP bond. To test this, you need to run
       multiple commands on the target host and on other servers in parallel.
     </para>
     <para>
       For example, on host <literal>server1</literal>, <command>iperf3 -s -p 5901</command>,
       and on client <literal>server2</literal>, <command>iperf3 -c -p 5902</command>.
     </para>
     <para>
       For more information on network interface bonding, see
       <link xlink:href="https://documentation.suse.com/sles/15-SP1/html/SLES-all/cha-network.html#sec-network-iface-bonding"/>.
     </para>
```

lproven: I found this paragraph very confusing. 

I have tried to clarify it by breaking it into several smaller [para]s but it is still not ideal. For instance, if I understand the original text correctly, it says that it is essential to run the test using more than just 2 servers, and then it specifically and explicitly gives an example that uses just 2 servers. 

I suggest seeking clarification from the original author of the hint.

asettle: This is their original text, and has been verified by 3 different people (see the attached issue report). I find it makes complete sense in the context.

asettle: I do, however, like the paragraph breakdown more than the blobbed text ;)

asettle: @ajaeger do you have any issue with this being broken down like this?

ajaeger: Breaking down helps IMHO - thanks!

PR #288, opened 2020-04-30
Creating issue template with supported linking

asettle:


xml/MAIN.susestorage.xml
```diff
@@ -25,6 +25,7 @@
    <dm:bugtracker>
     <dm:url>https://github.com/SUSE/doc-ses/issues/new/choose</dm:url>
```
sknorr, line 26: You need to remove `/choose` from the URL because that is still incompatible with us submitting our own issue body.

xml/issue_template.md
```diff
@@ -0,0 +1,21 @@
+### Type of report
+   - [ ] bug
+   - [ ] feature request
+
+### Source Document
+@@source@@
+
+### Summary
+<!-- A clear and concise description of what the bug/feature request is. -->
```
sknorr, line 9: ftr, I think XML-type comments work as comments in Markdown but I am not 100% sure anymore.

asettle: Yeah, I had the same thought. I think they do. I just need to remember. I'll look it up properly. Consider it a placeholder.

PR #292, opened 2020-05-04
Updating salt commands to be on the minion prompt - Closes #291

### LINKED ISSUE #291: [doc] 6.2  Deploy Salt
6.2  Deploy Salt

https://susedoc.github.io/doc-ses/master/html/ses-deployment/deploy-cephadm.html#deploy-salt

Under `View each minion's fingerprint:` the terminal reads `root@master`. This is inconsistent with the line above that this needs to be executed on a minion.

Executing the example on the master does not give any output, executing this on the minion does indeed give me the desired and shown output.

asettle:


asettle:
> Although it is nice that you changed all of them, only the first one needs to be changed. All other commands need to be executed on the `salt-master`. But thanks for addressing this so fast!

No problem! I admittedly was going to check through first with Tomas, but thanks for verifying before we got there also ;) I'll switch back now.

asettle:
@SchoolGuy Just one thing to check though, did you see the note above this section?

```If the salt minion fingerprint comes back empty, make sure the salt minion; has a master; configuration and that it can communicate with the master.``` Did you verify this?

SchoolGuy:
@asettle Yes I did. I had the situation that I executed the command on my salt master (which was back then no minion) and thus the command came back empty. When executing it on a minion which was properly configured it came back with the corresponding fingerprint.

After I enabled the salt-minion on my salt-master, obviously the command also succeed with returning the fingerprint.

asettle:
> @asettle Yes I did. I had the situation that I executed the command on my salt master (which was back then no minion) and thus the command came back empty. When executing it on a minion which was properly configured it came back with the corresponding fingerprint.
> 
> After I enabled the salt-minion on my salt-master, obviously the command also succeed with returning the fingerprint.

Thanks for verifying :)

PR #342, opened 2020-06-03
Explictly mention the mons IP usage when mount cephfs

### LINKED ISSUE #341: SES7 admin guide: explictly mention the mons IP usage when mount cephfs
Section "14.1.1 Client Preparation" in this document already recommends the installation of ceph-common package to use the full mount syntax when mounting a ceph filesystem.

It's worth explicitly mentioning that, without this package (and thus without the mount.ceph helper), the monitors IPs will need to be used instead of their names, as the kernel client won't be doing name resolution.

asettle:
Closes #341

### DIFF
--- a/xml/admin_ceph_cephfs.xml
+++ b/xml/admin_ceph_cephfs.xml
@@ -50,6 +50,20 @@
     <package>ceph-common</package> package (which is shipped with &sle;) should
     be installed before trying to mount &cephfs;.
    </para>
+   <important>
+     <para>
+       Without the <package>ceph-common</package> package (and thus without the
+       <command>mount.ceph</command> helper), the monitors IPs will need to
+       be used instead of their names, as the kernel client will be unable to
+       perform name resolution.
+     </para>
+     <para>
+       The basic mount syntax is:
+     </para>
+<screen>
+&prompt.cephuser; sudo mount -t ceph monip[:port][,monip2[:port]...]:/[subdir] mnt
+</screen>
+   </important>
   </sect2>
 
   <sect2 xml:id="Creating-Secret-File">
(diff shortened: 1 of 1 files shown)


xml/admin_ceph_cephfs.xml
```diff
@@ -50,6 +50,20 @@
...
+       Without the <package>ceph-common</package> package (and thus without the
+       <command>mount.ceph</command> helper), the monitors IPs will need to
+       be used instead of their names, as the kernel client will be unable to
+       perform name resolution.
+     </para>
+     <para>
+       The basic mount syntax is:
+     </para>
+<screen>
+&prompt.cephuser; sudo mount -t ceph monip[:port][,monip2[:port]...]:/[subdir] mnt
```
tbazant, line 64: in SES, prompt entities already include the ending space ' '
```suggestion
&prompt.cephuser;sudo mount -t ceph monip[:port][,monip2[:port]...]:/[subdir] mnt
```

xml/admin_ceph_cephfs.xml
```diff
@@ -50,6 +50,20 @@
     <package>ceph-common</package> package (which is shipped with &sle;) should
     be installed before trying to mount &cephfs;.
    </para>
+   <important>
+     <para>
+       Without the <package>ceph-common</package> package (and thus without the
+       <command>mount.ceph</command> helper), the monitors IPs will need to
```
lproven, line 56:
```suggestion
       <command>mount.ceph</command> helper), the monitors' IPs will need to
```

lproven: Does "monitors" here mean MON nodes? 

If so, I would suggest:

... the MON nodes' IP addresses must be used instead of their names.

asettle: This is "monitors" :) https://www.kernel.org/doc/html/latest/filesystems/ceph.html#mount-syntax

PR #373, opened 2020-06-18
Replace "whitelist" and "blacklist" language

asettle:
Related to: https://github.com/ceph/ceph/pull/35628

### DIFF
--- a/xml/admin_install_salt.xml
+++ b/xml/admin_install_salt.xml
@@ -1098,8 +1098,8 @@ alternative_defaults:
     <title>Cluster Assignment</title>
     <para>
      In the <emphasis role="bold">cluster</emphasis> section you select minions
-     for your cluster. You can select all minions, or you can blacklist or
-     whitelist minions. Examples for a cluster called
+     for your cluster. You can select all minions, or you can blocklist or
+     ignorelist minions. Examples for a cluster called
      <emphasis role="bold">ceph</emphasis> follow.
     </para>
     <para>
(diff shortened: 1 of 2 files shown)


tbazant:
Iagree with Liam:
- to use 'ignorelist' as a verb is strange, a pure 'allow' is better
- to use 'ignorelist' as a noun has a wrong meaning, one can think that 'ignorelisted' members will be ignore, while the contrary is true; 'allow-list' sound better to me
- also 'blocklisting' something is weird, we're just 'blocking' it, aren't we?

lproven:
> Let's wait til our language stylish commitee decides about how to proceed with the politically correct terms. 

Yes, I think we must. I understand wanting to just get it done as it's a small change, but I feel we must be sure everyone's using the same terminology.

> And please let's use entities for this to be 'flexible' :-)

Good idea! I didn't think of that.

And not use a jokey name for the entity, either... ;-)

asettle:
I'm sorry @tbazant and @lproven but this isn't my decision, nor do I feel this is up for debate. As I've linked in the commit message https://github.com/ceph/ceph/pull/35628 - this is the decision being made upstream. We need to reflect that to ensure that we are consistent. I'm keeping an eye on any debate or questions made upstream before even remotely merging this.

asettle:
@lproven @tbazant - just FYI though, I do agree with the fact that ignorelist and blocklist are not great terms. Can I take your comments and put them on the original PR?

asettle:
Accidental merge - will revert shortly.

Issue #382, opened 2020-06-22
[doc] 6.3.2.7  Configure Time Server - how to set time_server to disabled?


Martin-Weiss:
### Type of report
   - [ x] bug
   - [ ] feature request

### Source Document
6.3.2.7  Configure Time Server:

https://susedoc.github.io/doc-ses/master/single-html/ses-deployment/#deploy-cephadm-configure-ntp

### Summary
It seems that the default setting we deliver is this:

  o- time_server ..................................................................................... [enabled, no server host set]
    o- external_servers .................................................................................................... [empty]
    o- server_hostname ................................................................................................... [not set]
    o- subnet ............................................................................................................ [not set]

Because the time synchronization is already setup - we do not want the "time_server" to be configured via ceph-salt so we want to change "enabled" to "disabled".

How can this be done?

### Notes
Could we add the explanation on how to disable the time_server configuration via ceph-salt?

Issue #381, opened 2020-06-22
[doc] 6.3.2.3  Specify Salt Minions Managed by <code class="systemitem">cephadm</code>


Martin-Weiss:
### Type of report
   - [x] bug
   - [ ] feature request

### Source Document
6.3.2.3  Specify Salt Minions Managed by <code class="systemitem">cephadm</code>:

https://susedoc.github.io/doc-ses/master/single-html/ses-deployment/#deploy-cephadm-configure-cephadm

### Summary
At this stage of reading it is unclear for which minions the role "cephadm" should be added. All? only OSDs? Gateways?

### Notes
Could we get the information added if we should add all or some / which ones?

Issue #358, opened 2020-06-15
[doc] 6.3.2.7  Configure Time Server


toabctl:
### Type of report
   - [ ] bug
   - [ X ] feature request

### Source Document
6.3.2.7  Configure Time Server:

https://susedoc.github.io/doc-ses/master/html/ses-deployment/deploy-cephadm.html#deploy-cephadm-configure-ntp

### Summary
In the time-server docs, it should be mentioned that it's also possible to disable time server handling (in case the nodes already have ntp configured).

```
/time_server> ls
o- time_server .......................................................................................................... [disabled]
  o- external_servers ...................................................................................................... [empty]
  o- server_hostname ..................................................................................................... [not set]
  o- subnet .............................................................................................................. [not set]
/time_server>
```

asettle:
@toabctl where would I be able to find the instructions on how to disable time server handling?

toabctl:
@asettle I found out by trying (which was not difficult given that I've seen the PR on github implementing this). But I think it would be good to mention that you can disable `chrony` configuration if you already have time management (via chrony or ntp or something) available on the nodes.

asettle:
> @asettle I found out by trying (which was not difficult given that I've seen the PR on github implementing this). But I think it would be good to mention that you can disable `chrony` configuration if you already have time management (via chrony or ntp or something) available on the nodes.

For the sake of tracking, do you have a link to the PR? :)

toabctl:
https://github.com/ceph/ceph-salt/commit/a87c5be4ae0c4fb14d2462afc5b53cbd43aa0423

PR #392, opened 2020-06-26
Multiple updates to the cephadm deployment procedure

### LINKED ISSUE #384: [doc] 6.5.2.3  Applying Cluster Specification - structure for mon and mgr in one cluster.yaml?
### Type of report
   - [x] bug
   - [ ] feature request

### Source Document
6.5.2.3  Applying Cluster Specification:

https://susedoc.github.io/doc-ses/master/single-html/ses-deployment/#cephadm-appply-cluster-specs

### Summary
With MON and MGR in one cluster.yaml only the MGRs get deployed

### Steps To Reproduce
I tried to follow this and created a cluster.yaml this way:

```
service_type: mon
placement:
 hosts:
  - ses7-01
  - ses7-02
  - ses7-03
service_type: mgr
placement:
 hosts:
  - ses7-01
  - ses7-02
  - ses7-03
```

Then I  applied this with `ceph orch apply -i cluster.yaml`

### Expected Results
Expectation was to get three MGR and three MONs

### Actual Results
Just had 1 MON but three MGR

### Notes
My workaround was then to create two yaml files: 
mon.yaml and mgr.yaml which had only one of them and then I applied them separately.

So question is how can we configure a single cluster.yaml with all services included - or is the concept to have one per component?

tbazant:
Closes #384 #382 #358 #381

### DIFF
--- a/xml/deploy_cephadm.xml
+++ b/xml/deploy_cephadm.xml
@@ -548,11 +548,11 @@ o- ssh ................................................... [Key Pair set]
      </listitem>
      <listitem>
       <para>
-       If your site already has a single source of time, specify the external
+       If your site already has a single source of time, specify the
        host name of the time source:
       </para>
 <screen>
- &prompt.smaster;ceph-salt config /time_server/server_hostname set <replaceable>pool.ntp.org</replaceable>
+ &prompt.smaster;ceph-salt config /time_server/server_hostname set <replaceable>time-server.example.com</replaceable>
 </screen>
      </listitem>
      <listitem>
--- a/xml/deploy_cephadm.xml
+++ b/xml/deploy_cephadm.xml
@@ -316,6 +316,7 @@ o- / ............................................................... [...]
   | o- advanced .................................................... [...]
   | o- ceph_conf ................................................... [...]
   | o- dashboard ................................................... [...]
+  |   o- force_password_update ................................. [enabled]
   |   o- password ................................... [randomly generated]
   |   o- username ................................................ [admin]
   | o- mon_ip ............................................. [10.100.25.51]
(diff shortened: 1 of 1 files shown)


xml/deploy_cephadm.xml
```diff
@@ -531,24 +533,51 @@ o- ssh ................................................... [Key Pair set]
...
+&prompt.smaster;ceph-salt config /time_server disable
+</screen>
+     </listitem>
+     <listitem>
+      <para>
+       If your site already has a single source of time, specify the external
+       host name of the time source:
+      </para>
+<screen>
+ &prompt.smaster;ceph-salt config /time_server/server_hostname set <replaceable>pool.ntp.org</replaceable>
```
ricardoasmarques, line 555: `pool.ntp.org` is already being used for "external time servers", maybe we should use something that makes it clear that this is an "internal" host?

xml/deploy_cephadm.xml
```diff
@@ -531,24 +533,51 @@ o- ssh ................................................... [Key Pair set]
...
+       If all cluster hosts are already configured to synchronize their time
+       using a NTP service of choice, disable time server handling completely:
+      </para>
+<screen>
+&prompt.smaster;ceph-salt config /time_server disable
+</screen>
+     </listitem>
+     <listitem>
+      <para>
+       If your site already has a single source of time, specify the external
```
ricardoasmarques, line 551: I think we should remove the `external` word here.

xml/deploy_cephadm.xml
```diff
@@ -441,8 +442,9 @@ o- minions ................................................. [Minions: 5]
    <sect3 xml:id="deploy-cephadm-configure-cephadm">
     <title>Specify &sminion;s Managed by &cephadm;</title>
     <para>
-     Specify which minions will belong to the &ceph; cluster and will be
-     managed by &cephadm;:
+     Specify which hosts will belong to the &ceph; cluster and will be managed
+     by &cephadm;. Include all hosts that will run &ceph; services as well as
+     the &adm;:
```
ricardoasmarques, line 447: `as well as the &adm;` -> Just for reference, this will likely change with https://github.com/ceph/ceph-salt/issues/252 but for now, that's correct and we should keep this sentece.

xml/deploy_cephadm.xml
```diff
@@ -758,6 +796,7 @@ o- / ............................................................... [...]
   | o- advanced .................................................... [...]
   | o- ceph_conf ................................................... [...]
   | o- dashboard ................................................... [...]
+  |   o- force_password_update ................................. [enabled]
```
LenzGr, line 799: Note that I also update this screen output in #409 - we also added two more options to configure the dashboard's SSL certificate and key via `ceph-salt`.

asettle: We'll deal with the merge conflict when it comes to it. It shouldn't be much. I just need @sebastian-philipp to review this PR, and then we can merge.

PR #406, opened 2020-06-30
Adding warning to ceph-salt export command  - Closes#339

asettle:
Originally the request was to add umask, but to avoid overcomplication, adding a warning to the export command should ensure individual are aware of the risk.

### DIFF
--- a/xml/deploy_cephadm.xml
+++ b/xml/deploy_cephadm.xml
@@ -529,7 +529,7 @@ o- ssh ................................................... [Key Pair set]
 </screen>
    <note>
      <para>
-       The resulting exported json configuration includes a private SSH key.
+       The resulting exported JSON configuration includes a private SSH key.
        Take care when dumping this file to ensure that it can not be accessed
        by others.
      </para>
--- a/xml/deploy_cephadm.xml
+++ b/xml/deploy_cephadm.xml
@@ -799,9 +799,9 @@ config: OK
 <screen>&prompt.smaster;ceph-salt export > cluster.json</screen>
     <warning>
       <para>
-       Executing the <command>ceph-salt export</command> includes the
+       The output of the <command>ceph-salt export</command> includes the
        SSH private key. If you are concerned about the security implications,
-       then do not execute this command unless necessary.
+       do not execute this command without taking appropriate precautions.
       </para>
     </warning>
     <para>
--- a/xml/deploy_cephadm.xml
+++ b/xml/deploy_cephadm.xml
@@ -527,6 +527,23 @@ o- ssh ................................................... [Key Pair set]
   o- private_key ...... [53:b1:eb:65:d2:3a:ff:51:6c:e2:1b:ca:84:8e:0e:83]
   o- public_key ....... [53:b1:eb:65:d2:3a:ff:51:6c:e2:1b:ca:84:8e:0e:83]
 </screen>
+   <note>
+     <para>
+       The resulting exported json configuration includes a private SSH key.
+       Take care when dumping this file to ensure that it can not be accessed
+       by others.
+     </para>
+     <para>
+       To ensure this does not occur, you can take an extra precaution by
+       using <command>umask</command>:
+     </para>
+<screen>&prompt.cephuser;umask 077
+&prompt.cephuser;ceph-salt export > cluster.json</screen>
+     <para>
+       The <command>umask</command> change avoids accidental granting of
+       non-root access on the same host.
+     </para>
+   </note>
    </sect3>
    <sect3 xml:id="deploy-cephadm-configure-ntp">
     <title>Configure Time Server</title>
--- a/xml/deploy_cephadm.xml
+++ b/xml/deploy_cephadm.xml
@@ -527,23 +527,6 @@ o- ssh ................................................... [Key Pair set]
   o- private_key ...... [53:b1:eb:65:d2:3a:ff:51:6c:e2:1b:ca:84:8e:0e:83]
   o- public_key ....... [53:b1:eb:65:d2:3a:ff:51:6c:e2:1b:ca:84:8e:0e:83]
 </screen>
-   <note>
-     <para>
-       The resulting exported JSON configuration includes a private SSH key.
-       Take care when dumping this file to ensure that it can not be accessed
-       by others.
-     </para>
-     <para>
-       To ensure this does not occur, you can take an extra precaution by
-       using <command>umask</command>:
-     </para>
-<screen>&prompt.cephuser;umask 077
-&prompt.cephuser;ceph-salt export > cluster.json</screen>
-     <para>
-       The <command>umask</command> change avoids accidental granting of
-       non-root access on the same host.
-     </para>
-   </note>
    </sect3>
    <sect3 xml:id="deploy-cephadm-configure-ntp">
     <title>Configure Time Server</title>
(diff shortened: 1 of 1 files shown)


xml/deploy_cephadm.xml
```diff
@@ -527,6 +527,23 @@ o- ssh ................................................... [Key Pair set]
   o- private_key ...... [53:b1:eb:65:d2:3a:ff:51:6c:e2:1b:ca:84:8e:0e:83]
   o- public_key ....... [53:b1:eb:65:d2:3a:ff:51:6c:e2:1b:ca:84:8e:0e:83]
 </screen>
+   <note>
+     <para>
+       The resulting exported json configuration includes a private SSH key.
```
lproven, line 532:
```suggestion
       The resulting exported JSON configuration includes a private SSH key.
```

xml/deploy_cephadm.xml
```diff
@@ -527,6 +527,23 @@ o- ssh ................................................... [Key Pair set]
...
+     <para>
+       The resulting exported JSON configuration includes a private SSH key.
+       Take care when dumping this file to ensure that it can not be accessed
+       by others.
+     </para>
+     <para>
+       To ensure this does not occur, you can take an extra precaution by
+       using <command>umask</command>:
+     </para>
+<screen>&prompt.cephuser;umask 077
```
denisok, line 540: this would change mask for all further commands as well.
So we either need to instruct customer to first get current mask with `umask -p` and then restore.
Or execute something like: `(umask 077 && ceph-salt export > cluster.json)` instead.

LenzGr: Yes, the latter option would make more sense to me as well. Or how about just using a `chmod 600 cluster.json` right after the file was exported?
```
ceph-salt export > cluster.json && chmod 600 ./cluster.json
```

denisok: in time between two commands, it is already intercepted by hackers :)

xml/deploy_cephadm.xml
```diff
@@ -527,6 +527,23 @@ o- ssh ................................................... [Key Pair set]
   o- private_key ...... [53:b1:eb:65:d2:3a:ff:51:6c:e2:1b:ca:84:8e:0e:83]
   o- public_key ....... [53:b1:eb:65:d2:3a:ff:51:6c:e2:1b:ca:84:8e:0e:83]
 </screen>
+   <note>
+     <para>
+       The resulting exported JSON configuration includes a private SSH key.
```
LenzGr, line 532: I am somewhat confused by how this paragraph begins. I could not find any mentions of `ceph-salt export` in this chapter before, so it feels strange to start with "the resulting exported JSON file" while not having talked about the option to export the configuration in the first place. I suggest to re-arrange this to explain that the config can be exported in JSON format using `ceph-salt export`, followed by an explanation why using `umask` is important here.

xml/deploy_cephadm.xml
LenzGr: Since my suggestion to use `chmod` leaves the file open for prying eyes for a (admittedly very short) period of time, I think Denis' suggestion of using `umask` in a subshell should be preferred:
```
(umask 077 && ceph-salt export > cluster.json)
```

xml/deploy_cephadm.xml
```diff
@@ -527,6 +527,23 @@ o- ssh ................................................... [Key Pair set]
...
+       by others.
+     </para>
+     <para>
+       To ensure this does not occur, you can take an extra precaution by
+       using <command>umask</command>:
+     </para>
+<screen>&prompt.cephuser;umask 077
+&prompt.cephuser;ceph-salt export > cluster.json</screen>
+     <para>
+       The <command>umask</command> change avoids accidental granting of
```
LenzGr, line 543: "Granting of non-root access" sounds weird. How about: "... avoids that the file's content can be read by non-root users."

xml/deploy_cephadm.xml
denisok: not that fast Lenz, I also has assumption that bash interpreter doesn't have any issues and those commands are executed in order )))
So strictly speaking priv key should be exported at all, or by special command and with special perm.

xml/deploy_cephadm.xml
asettle: Yeah it is repeating. I think it would be fine if we just removed this - thoughts?

xml/deploy_cephadm.xml
asettle: So, confusingly, it's explained here: https://susedoc.github.io/doc-ses/master/single-html/ses-deployment/#deploy-cephadm-configure-export which is _after_ this section (configure export is 6.3.2.13 and SSH gen is 6.3.2.6)
Perhaps the configure export section should be moved above the SSH? Or, the umask note should be moved to the export chap... Thoughts?

xml/deploy_cephadm.xml
LenzGr: Wouldn't we enforce the order by using a semicolon instead?
```
(umask 077 ;  ceph-salt export > cluster.json)
```
I think the question of including the SSH key is out of scope for this PR, let's discuss this in https://github.com/ceph/ceph-salt/issues/288 instead.

xml/deploy_cephadm.xml
LenzGr: Sounds good to me!

xml/deploy_cephadm.xml
denisok: yes, `&&` also should work, if we assume all underling system works.
But really what I am saying is that we are looking for workarounds instead of fixing the issue,

denisok: we need bash expert. @smithfarm this thread lacks some comments.

xml/deploy_cephadm.xml
LenzGr: I think it's fine to conclude the configuration chapter with a note that the configuration can be exported for backup purposes, so the order of the section should stay. But I agree, I would move the umask note to the export part instead.

asettle: Works for me.

xml/deploy_cephadm.xml
smithfarm: I don't think it's a good idea for our documentation to try to teach customers Unix Fundamentals. My opinion is: `umask` is too technical, and we should not mention it here at all.

LenzGr: OK, thanks for your suggestion. So in a nutshell, let's stick to `ceph-salt export > cluster.json` and add a note that explains that this is inherently unsafe.

denisok: +1

xml/deploy_cephadm.xml
```diff
@@ -797,6 +797,13 @@ config: OK
      valid, it is a good idea to export its configuration to a file:
     </para>
 <screen>&prompt.smaster;ceph-salt export > cluster.json</screen>
+    <warning>
+      <para>
+       Executing the <command>ceph-salt export</command> includes the
```
smithfarm, line 802: "Executing the `ceph-salt export` command includes the SSH private key" doesn't sound quite right.

How about: "The output of the `ceph-salt export` command includes the SSH private key" ?

xml/deploy_cephadm.xml
```diff
@@ -797,6 +797,13 @@ config: OK
...
+       SSH private key. If you are concerned about the security implications,
+       then do not execute this command unless necessary.
```
smithfarm, line 804: instead of "do not execute this command unless necessary" we could say "do not execute this command without taking appropriate precautions" ?

PR #415, opened 2020-07-08
Remove LTSS repositories before migration (bsc#1173306)

tbazant:
LTSS repos break migration form SES5

### DIFF
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -482,7 +482,7 @@ mds standby for name = mds.456-another-mds
   <sect2 xml:id="upgrade-prepare-repos">
    <title>Required Software Repositories</title>
    <para>
-    Check that required repositories are configured on each cluster's node. To
+    Check that required repositories are configured on each host of the cluster. To
     list all available repositories, run
    </para>
 <screen>
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -491,7 +491,7 @@ mds standby for name = mds.456-another-mds
 <important>
  <title>Remove &productname; &prevproductnumber; LTSS Repositories</title>
  <para>
-  Upgrade will fail if LTSS repositories are present in the &productname; &prevproductnumber;.
+  Upgrades will fail if LTSS repositories are configured in &productname; &prevproductnumber;.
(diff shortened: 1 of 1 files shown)


xml/admin_ceph_upgrade.xml
```diff
@@ -485,6 +485,24 @@ mds standby for name = mds.456-another-mds
     Check that required repositories are configured on each cluster's node. To
```
lproven, line 485: > each cluster's node

...(?)

What does this use of node in the singular mean? 1 node in each? All nodes in each cluster seems more likely to me, but I am not sure. If so, how about "each cluster's nodes"?

xml/admin_ceph_upgrade.xml
```diff
@@ -485,6 +485,24 @@ mds standby for name = mds.456-another-mds
...
     list all available repositories, run
    </para>
+<screen>
+&prompt.sminion;zypper lr
+</screen>
+<important>
+ <title>Remove &productname; &prevproductnumber; LTSS Repositories</title>
+ <para>
+  Upgrade will fail if LTSS repositories are present in the &productname; &prevproductnumber;.
+  Find their ID and remove them from the system, for example:
```
lproven, line 495:
```suggestion
  Find their IDs and remove them from the system; for example:
```

xml/admin_ceph_upgrade.xml
```diff
@@ -485,6 +485,24 @@ mds standby for name = mds.456-another-mds
     Check that required repositories are configured on each cluster's node. To
     list all available repositories, run
    </para>
+<screen>
+&prompt.sminion;zypper lr
+</screen>
+<important>
+ <title>Remove &productname; &prevproductnumber; LTSS Repositories</title>
+ <para>
+  Upgrade will fail if LTSS repositories are present in the &productname; &prevproductnumber;.
```
lproven, line 494:
```suggestion
  Upgrades will fail if LTSS repositories are configured in &productname; &prevproductnumber;.
```

xml/admin_ceph_upgrade.xml
tbazant, line 485:
```suggestion
    Check that required repositories are configured on each host of the cluster. To
```

tbazant: fixed

PR #424, opened 2020-07-20
Updating time sync disable note with xref (BSC#1171596)

asettle:
Fujistu requested a clarification on the disabling time sync note. See https://bugzilla.suse.com/show_bug.cgi?id=1171596#c19

### DIFF
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -361,7 +361,8 @@ mds standby for name = mds.456-another-mds
     <para>
       If you want to manually maintain the <systemitem class="daemon">chronyd</systemitem>
       configuration, follow the instructions below and ensure you disable
-      the time configuration (as describe below).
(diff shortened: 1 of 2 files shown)


xml/admin_ceph_upgrade.xml
```diff
@@ -361,7 +361,8 @@ mds standby for name = mds.456-another-mds
     <para>
       If you want to manually maintain the <systemitem class="daemon">chronyd</systemitem>
       configuration, follow the instructions below and ensure you disable
-      the time configuration (as describe below).
+      the time configuration. See <xref linkend="disable-time-sync"/> for more
```
lproven, line 364:
```suggestion
      <command>ntpd</command> time configuration. See <xref linkend="disable-time-sync"/> for more
```

lproven: I am not sure [command] is the best tag. [literal] or something else might be fine too.

PR #436, opened 2020-07-22
Updating networking section - Closes #363

### LINKED ISSUE #363: Document how to configure multiple networks (eg. cluster/public/gateway networks) with SES 7
Similar to https://documentation.suse.com/ses/6/single-html/ses-deployment/#network-overview (for SES 6), we need documentation how different networks can be used for different parts of ceph.

asettle:
Still need to update section with adding cluster and gateway networks


xml/deployment_hwrecommend.xml
```diff
@@ -33,13 +33,13 @@
   <itemizedlist>
    <listitem>
     <para>
-     A trusted internal network, the back-end network called the
-     <literal>cluster network</literal>.
+     A public client network called <literal>public network</literal>.
```
smithfarm, line 36: `A front-end network called the <literal>public network</literal>.`

xml/deployment_hwrecommend.xml
```diff
@@ -51,7 +51,7 @@
 
   <para>
    The trusted internal network is the back-end network between the OSD nodes
```
smithfarm, line 53: The cluster network is the back-end network . . .

xml/deployment_hwrecommend.xml
```diff
@@ -51,7 +51,7 @@
...
-   for replication, re-balancing and recovery.Ideally, this network provides
+   for replication, re-balancing and recovery. Ideally, this network provides
```
smithfarm, line 54: If configured, this optional network would ideally provide

xml/deployment_hwrecommend.xml
```diff
@@ -71,12 +71,20 @@
...
-    network interfaces provides aggregation and fault-tolerance. The bond
-    should then be used to provide two VLAN interfaces, one for the public
-    network, and the second for the cluster network. Details on bonding the
-    interfaces can be found in
+     We recommend a single fault-tolerant network with enough bandwidth to
+     fulfil your requirements. For the &ceph; public network environment, we
+     recommend two bonded 25 GbE (or
+     faster) network interfaces bonded using 802.3ad (LACP). This is considered
+     the minimal setup for &ceph;. If you are also using a cluster network, we
+     recommend four bonded 25 GbE. Doubling the network interfaces provides
```
smithfarm, line 79: four bonded 25 GbE network interfaces.

xml/deployment_hwrecommend.xml
```diff
@@ -33,27 +33,28 @@
...
     </para>
    </listitem>
    <listitem>
     <para>
-     Client networks for gateways, these are optional.
+     One or more client networks for gateways. This is optional and
+     beyond the scope of this chapter.
     </para>
    </listitem>
   </itemizedlist>
```
smithfarm, line 52: Here we are missing a paragraph describing the public network. Maybe this one:

"The public network is the network over which Ceph daemons communicate with each other and with their clients. That means: all Ceph cluster traffic goes over this network except in the case when a cluster network is configured (see the next paragraph)."

asettle: Depends. Are we assuming the user doesn't understand what a public network is vs. a cluster network? In that case, we should probably add them as glossary definition and cross-ref that, rather than insert definitions :thinking:

smithfarm: Isn't is weird that there is a paragraph describing what the cluster network is, then? Either have paragraphs describing both public and cluster, or no paragraphs describing either.

asettle: It is, to be honest. It stands out like a sore thumb. What I meant by my comment above is that the cluster network definition is a standout, and an awkward one. We can define both, but my recommendation would be for both to go into the glossary. What are your thoughts, @tbazant ?

xml/deployment_hwrecommend.xml
```diff
@@ -33,27 +33,28 @@
...
 
   <para>
-   The trusted internal network is the back-end network between the OSD nodes
-   for replication, re-balancing and recovery.Ideally, this network provides
-   twice the bandwidth of the public network with default 3-way replication
-   since the primary OSD sends 2 copies to other OSDs via this network. The
+   The cluster network is the back-end network between the OSD nodes
```
lproven, line 54:
```suggestion
   The cluster network is the back-end network between the OSD nodes,
```

xml/deployment_hwrecommend.xml
```diff
@@ -33,27 +33,28 @@
...
+   for replication, re-balancing and recovery. If configured, this optional
```
lproven, line 55:
```suggestion
   for replication, re-balancing, and recovery. If configured, this optional
```

xml/deployment_hwrecommend.xml
```diff
@@ -33,27 +33,28 @@
...
+   network would ideally provide twice the bandwidth of the public network with
+   default 3-way replication since the primary OSD sends 2 copies to other OSDs via this network. The
```
lproven, line 57:
```suggestion
   default 3-way replication, since the primary OSD sends two copies to other OSDs
   via this network. The
```

xml/deployment_hwrecommend.xml
```diff
@@ -71,12 +72,20 @@
...
    <para>
-    For the &ceph; network environment, we recommend two bonded 25 GbE (or
-    faster) network interfaces bonded using 802.3ad (LACP). The use of two
-    network interfaces provides aggregation and fault-tolerance. The bond
-    should then be used to provide two VLAN interfaces, one for the public
-    network, and the second for the cluster network. Details on bonding the
-    interfaces can be found in
+     We recommend a single fault-tolerant network with enough bandwidth to
+     fulfil your requirements. For the &ceph; public network environment, we
+     recommend two bonded 25 GbE (or
```
lproven, line 77:
```suggestion
     recommend two bonded 25&nbsp;GbE (or
```

xml/deployment_hwrecommend.xml
```diff
@@ -71,12 +72,20 @@
...
+     faster) network interfaces bonded using 802.3ad (LACP). This is considered
+     the minimal setup for &ceph;. If you are also using a cluster network, we
+     recommend four bonded 25 GbE network interfaces. Doubling the network interfaces
```
lproven, line 80:
```suggestion
     recommend four bonded 25&nbsp;GbE network interfaces. Doubling the network interfaces
```

xml/deployment_hwrecommend.xml
```diff
@@ -33,27 +33,29 @@
...
     <para>
-     A trusted internal network, the back-end network called the
-     <literal>cluster network</literal>.
+     A front-end network called the <literal>public network</literal>.
     </para>
    </listitem>
    <listitem>
     <para>
-     A public client network called <literal>public network</literal>.
+     A trusted internal network, the back-end network called the
```
lproven, line 41:
```suggestion
     A trusted internal network, the back-end network, called the
```

xml/deployment_hwrecommend.xml
```diff
@@ -33,27 +33,29 @@
...
 
   <para>
-   The trusted internal network is the back-end network between the OSD nodes
-   for replication, re-balancing and recovery.Ideally, this network provides
-   twice the bandwidth of the public network with default 3-way replication
-   since the primary OSD sends 2 copies to other OSDs via this network. The
+   The cluster network is the back-end network between the OSD nodes,
+   for replication, re-balancing, and recovery. If configured, this optional
+   network would ideally provide twice the bandwidth of the public network with
+   default 3-way replication, since the primary OSD sends two copies to other OSDs
```
lproven, line 57:
```suggestion
   default three-way replication, since the primary OSD sends two copies to other OSDs
```

xml/deployment_hwrecommend.xml
```diff
@@ -178,18 +188,10 @@
...
-<screen>[mon.MON1]
- public network = 192.168.123.0/24
-
- [mon.MON2]
- public network = 1.2.3.0/24
-
- [mon.MON3]
- public network = 242.12.33.12/0</screen>
+<screen>[global]
+ public network = 192.168.123.0/24, 1.2.3.0/24, 242.12.33.12/0</screen>
```
lproven, line 194: Question.
The 1st example is in the public Class C address range (192.168.0.1–192.168.254.254, from memory). 

If we are using these as examples, then should the others **also** be public? 10.x.y.z & 176.16.y.z?

If not & we want real ranges... SUSE's ranges?

1.2.3.0 is a real address range and it's not ours. 242.x.y.z I don't know.

lproven: I think @asettle is away for a while now, so @smithfarm I would really appreciate your opinion on this.

smithfarm:
@lproven Those IP addresses are already in the document, so I don't see any reason to change them. I don't think any reader would interpret this as meaning that they should **literally** write `1.2.3.0/24` in their ceph.conf.

The point is that each monitor can be on a different network, and the cluster will still work provided:

(a) the packets are routed properly (routing being beyond the scope of this chapter) **and** 
(b) the networks are all mentioned in the `public_network` setting in ceph.conf, as illustrated here.

Technically, it matters not what the network numbers are: the Linux TCP/IP stack doesn't know (or care) about the difference between "public" and "non-public" ranges.

xml/deployment_hwrecommend.xml
asettle: After reviewing this again, decided to go with both.

xml/deployment_hwrecommend.xml
```diff
@@ -71,12 +79,20 @@
...
-    should then be used to provide two VLAN interfaces, one for the public
-    network, and the second for the cluster network. Details on bonding the
-    interfaces can be found in
+     We recommend a single fault-tolerant network with enough bandwidth to
+     fulfil your requirements. For the &ceph; public network environment, we
+     recommend two bonded 25&nbsp;GbE (or
+     faster) network interfaces bonded using 802.3ad (LACP). This is considered
+     the minimal setup for &ceph;. If you are also using a cluster network, we
+     recommend four bonded 25&nbsp;GbE network interfaces. Doubling the network interfaces
+     provides aggregation and fault-tolerance.
```
jan--f, line 88: This might be nit-picking but I stumbled somewhat over that last sentence. Maybe being a bit more explicit here can add value?
```
Bonding two or more network interfaces provides better throughput via link aggregation and, given redundant links and switches improved fault tolerance and maintainability.
```

PR #499, opened 2020-08-27
Updating note about rook manifest location - Closes#490

asettle:


asettle:
@lproven I actually hate the way I phrased this. Suggestion box open.

BlaineEXE:
I'm not sure it needs to be a `<note>`, but I trust your judgment there.

lproven (changes requested):
Lines 116:
>  in the 'rook-ceph' namespace. If you need to change it, adopt all

**adapt** all

xml/deploy_caasp.xml
```diff
@@ -86,9 +86,9 @@ Ignore "Ceph Flexvolume Configuration" section -->
   <para>
    The &rook; orchestrator uses configuration files in YAML format called
    <emphasis>manifests</emphasis>. The manifests you need are included in the
-   <package>rook-k8s-yaml</package> RPM package. Install it by running
+   <package>rook-k8s-yaml</package> RPM package. These are automatically installed
```
BlaineEXE, line 89: weak suggestion:

```suggestion
   <package>rook-k8s-yaml</package> RPM package. These are installed
```

lproven (commented):
Lines 105, 106, 116 etc., e.g.

> in the 'rook-ceph' namespace. If you need to change it, adopt all

Should we not be using some markup rather than single-quotes? I am not sure, but I suspect so.

lproven (commented):
Line 70:

> &casp; 4 worker nodes with a number of extra disks attached as storage for

Just for clarity: is this a CaaSP version number, or is it how many nodes you need? If it is a count of nodes, then the word "four" instead of the digit. I would prefer less-ambiguous phrasing such as:

 - at least four CaaSP worker nodes with attached disks

asettle:
> Line 70:
> 
> > &casp; 4 worker nodes with a number of extra disks attached as storage for
> 
> Just for clarity: is this a CaaSP version number, or is it how many nodes you need? If it is a count of nodes, then the word "four" instead of the digit. I would prefer less-ambiguous phrasing such as:
> 
> * at least four CaaSP worker nodes with attached disks

We can probably leave this until we actually publish and edit this document :) lets keep this in mind until later, this is out of scope for a small bug report.

lproven (approved):

asettle:
@lproven for the record though, something went a bit funky in edit town but originally it would have said `At least three SUSE CaaS Platform worker nodes, with at least one additional disk attached to each worker node as storage for the OSD. We recommend four SUSE CaaS Platform worker nodes.` - so, something to keep in mind there. Ref: https://documentation.suse.com/ses/6/single-html/ses-deployment/#kube-prereq

lproven:
> something went a bit funky in edit town but originally it would have said[...]

Ahaa! That makes a lot more sense.

PR #503, opened 2020-08-28
Updating upgrade of CTDB clusters (BSC#1144917)

asettle:


ddiss (changes requested):
@asettle Thanks a lot for putting this together. I've pushed some minor changes to https://github.com/SUSE/doc-ses/tree/asettle-ddiss-sambaupgrade-1144917 .
If your okay with them then they can be squashed in and merged.

PR #518, opened 2020-09-08
Update deploy_cephadm.xml

lezare:
The option for the command is "--service-type" and not just "--type", thus the current wording is confusing.


tbazant (approved):
LGTM, thanks

PR #565, opened 2020-09-14
admin_nfsganesha.xml: update to replace usage of `showmount`

mgfritch:
NFSv4 does not advertise exports via rpcbind/mountd. Instead these can
be enumerated by a client from the root of the pseudo NFSv4 filesystem

Signed-off-by: Michael Fritch <mfritch@suse.com>


tbazant (changes requested):
Jusr few minor fixes.

asettle (approved):

tbazant (approved):
lgtm

PR #566, opened 2020-09-15
admin_saltcluster.xml: fix "encryption:" tag

skouya:
"encryption:" tag in DriveGroups definition is wrong. "encrypted:" is correct.


tbazant (approved):
lgtm

PR #588, opened 2020-09-21
Update admin_ceph_cephfs.xml

luis-henrix:
A few documentation suggestions:

- "13.1.1 Client Preparation": there was no mention to SLE15* clients, so just
  use 'or later' instead of enumerating every single release.

- "13.6.1 Limitations": mention that SLE12-SP3 already has the required
  backports to handle cephfs quotas

- "13.7.1 Creating Snapshots": the kernel clients limitation to 400 snapshots is
  per-filesystem.  Also promoted this info to an 'important' box.


asettle (approved):

jan--f (approved):
:+1:

PR #602, opened 2020-09-24
Adding troubleshooting docs for iscsi

asettle:


xml/bp_troubleshooting_iscsi.xml
```diff
@@ -13,7 +13,10 @@
    <dm:release>SES 7</dm:release>
   </dm:docmanager>
  </info>
- <para>
-  TBD
- </para>
+ <sect1 xml:id="troubleshooting-debug-lrbd">
+   <title>Debugging Incompability of <literal>lrbd</literal></title>
+   <para>
+      ceph-iscsi does not support TPG authentication alongside initiator ACLs.
```
ricardoasmarques, line 19: This information is only relevant during the upgrade from `SES5` to `SES6`. Do we have a section for this upgrade?

asettle: @ricardoasmarques :thinking: we'd literally just have to include it in the upgrade docs for SES 6, suitable?

PR #608, opened 2020-09-28
Updating ceph cluster shutdown commands

### LINKED ISSUE #584: 5.3 Shutdown and Restart of the Whole Ceph Cluster in SES7
**Link To Section**
SES6: https://susedoc.github.io/doc-ses/ses6/html/ses-admin/ceph-operating-services.html#ceph-cluster-shutdown
SES7: https://susedoc.github.io/doc-ses/master/html/ses-admin/cha-ceph-operating.html#ceph-cluster-shutdown

**Summary**
`"Shutdown and Restart of the Whole Ceph Cluster"` section must be updated for cephadm/containers environment.

**Steps To Reproduce**
In SES6, we use a Deepsea runner to guarantee that services are stoped in the correct order:
1. gateways
2. osds
3. mons/mgrs

But we don't guarantee that on SES7 docs.

**Expected Results**
We need to check if current documentation is still valid for `SES7`.

From the links below, the workflow should now be something like:

_1) Get cluster FSID:_
```
# ceph -s
```

_2) Set some OSD flags:_
```
# ceph osd set noout
# ceph osd set nobackfill
# ceph osd set norecover
# ceph osd set norebalance
# ceph osd set nodown
# ceph osd set pause
```

_3) Stop ceph service on all nodes:_
```
# salt TARGET cmd.run "systemctl stop ceph-<fsid>.target"
```

But we are still missing a way to guarantee that services are stopped in the correct order.

We may need a new `ceph-salt` or `cephadm` feature here.

_**(This workflow requires confirmation)**_

**Notes**
Some links that may contain useful information
- https://github.com/ceph/ceph-salt/issues/359
- https://jira.suse.com/browse/SES-2048
- https://ceph.io/planet/how-to-do-a-ceph-cluster-maintenance-shutdown/

asettle:
Closes #584

### DIFF
--- a/xml/admin_operating_services.xml
+++ b/xml/admin_operating_services.xml
@@ -260,13 +260,13 @@ osd.0                               ses-min2  running (67m)  3m ago
     <para>
      Stop all &ceph; services on all cluster nodes:
     </para>
-<screen>&prompt.smaster;salt <replaceable>TARGET</replaceable> cmd.run "systemctl stop ceph.target"</screen>
+<screen>&prompt.smaster;ceph-salt stop</screen>
    </step>
    <step>
     <para>
      Power off all cluster nodes:
     </para>
-<screen>&prompt.smaster;salt <replaceable>TARGET</replaceable> cmd.run "shutdown -h"</screen>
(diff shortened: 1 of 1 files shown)


ricardoasmarques (approved):
lgtm

PR #613, opened 2020-09-29
rename `registry.suse.de` -> `registry.suse.com`

mgfritch:
NFS example in admin guide should reference `registry.suse.com` instead
of `registry.suse.de`

Signed-off-by: Michael Fritch <mfritch@suse.com>


tbazant (approved):
lgtm, thanks

PR #616, opened 2020-09-29
Added section on using legacy kernel clients

luis-henrix:
Signed-off-by: Luis Henriques <lhenriques@suse.de>

### DIFF
--- a/xml/bp_troubleshooting_cephfs.xml
+++ b/xml/bp_troubleshooting_cephfs.xml
@@ -272,7 +272,7 @@ sudo zypper in ceph-common
    </sect2>
  </sect1>
  <sect1 xml:id="bp-troubleshooting-old-clients-mount">
-   <title>Mounting CephFS Using Old Kernel Clients</title>
+   <title>Mounting &cephfs; Using Old Kernel Clients</title>
    <para>
      The kernel in &sls; 15 SP2 includes a
      &cephfs; client that is able to take full advantage of all the features
--- a/xml/bp_troubleshooting_cephfs.xml
+++ b/xml/bp_troubleshooting_cephfs.xml
@@ -280,9 +280,9 @@ sudo zypper in ceph-common
      backported to this operating system.
    </para>
    <para>
-     However, it may be necessary to access &cephfs; from another systems that
+     However, it may be necessary to access &cephfs; from other systems that
      may provide an older &cephfs; client, which may not support all the
-     features required by an SES7 cluster. When this happens, the kernel client
+     features required by an &productname; &productnumber; cluster. When this happens, the kernel client
      will fail to mount the file system and will emit messages similar to the
      one shown below:
    </para>
@@ -641,8 +641,8 @@ sudo zypper in ceph-common
      A possible solution to allow an old kernel client to mount a recent
      &cephfs; is to modify the cluster CRUSH profile. CRUSH profiles define a
      set of CRUSH tunables that are named after the &ceph; versions in which they
-     were introduced. For example, the 'firefly' tunables are first supported in
-     the Firefly release (0.80) and older clients will not be able to access the
+     were introduced. For example, the <literal>firefly</literal> tunables are first supported in
+     the Firefly release (0.80), and older clients will not be able to access the
      cluster. Thus, to fix the problem shown above, the following command can be
      used:
    </para>
--- a/xml/bp_troubleshooting_cephfs.xml
+++ b/xml/bp_troubleshooting_cephfs.xml
@@ -640,7 +640,7 @@ sudo zypper in ceph-common
    <para>
      A possible solution to allow an old kernel client to mount a recent
      &cephfs; is to modify the cluster CRUSH profile. CRUSH profiles define a
-     set of CRUSH tunables that are named after the Ceph versions in which they
+     set of CRUSH tunables that are named after the &ceph; versions in which they
      were introduced. For example, the 'firefly' tunables are first supported in
      the Firefly release (0.80) and older clients will not be able to access the
      cluster. Thus, to fix the problem shown above, the following command can be
--- a/xml/bp_troubleshooting_cephfs.xml
+++ b/xml/bp_troubleshooting_cephfs.xml
@@ -274,10 +274,10 @@ sudo zypper in ceph-common
  <sect1 xml:id="bp-troubleshooting-old-clients-mount">
    <title>Mounting CephFS Using Old Kernel Clients</title>
    <para>
-     The Linux kernel available in the &sls; 15 SP2 Operating System includes a
+     The kernel in &sls; 15 SP2 includes a
      &cephfs; client that is able to take full advantage of all the features
      available on an SES7 cluster. All relevant features and bug fixes are
-     backported to this Operating System.
+     backported to this operating system.
    </para>
    <para>
      However, it may be necessary to access &cephfs; from another systems that
--- a/xml/bp_troubleshooting_cephfs.xml
+++ b/xml/bp_troubleshooting_cephfs.xml
@@ -271,4 +271,450 @@ sudo zypper in ceph-common
      </para>
    </sect2>
  </sect1>
+ <sect1 xml:id="bp-troubleshooting-old-clients-mount">
+   <title>Mounting CephFS Using Old Kernel Clients</title>
+   <para>
+     The Linux kernel available in the &sls; 15 SP2 Operating System includes a
+     &cephfs; client that is able to take full advantage of all the features
+     available on an SES7 cluster. All relevant features and bug fixes are
+     backported to this Operating System.
+   </para>
+   <para>
+     However, it may be necessary to access &cephfs; from another systems that
+     may provide an older &cephfs; client, which may not support all the
+     features required by an SES7 cluster. When this happens, the kernel client
+     will fail to mount the file system and will emit messages similar to the
+     one shown below:
+   </para>
+<screen>
+[ 4187.023633] libceph: mon0 192.168.122.150:6789 feature set mismatch, my 107b84a842aca &lt; server's 40107b84a842aca, missing 400000000000000
+[ 4187.023838] libceph: mon0 192.168.122.150:6789 missing required protocol features
+</screen>
+   <para>
+     The message above means that the MON identified 0x400000000000000 as the
+     missing feature in the client (the value 0x107b84a842aca represents all the
+     features supported by the client, while 0x40107b84a842aca represents the
+     minimum set of features required by the cluster). From the following
+     table, which shows the complete list of feature bits, we can see that the
+     missing feature bit 58 (2^58 = 0x400000000000000) is CRUSH_TUNABLES5,
+     NEW_OSDOPREPLY_ENCODING, or FS_FILE_LAYOUT_V2 (all these three features
+     share the same feature bit).
+   </para>
+   <table>
+     <title>Ceph Features</title>
+     <tgroup cols="3">
+       <colspec colnum="1" colname="col1" colwidth="*"/>
+       <colspec colnum="2" colname="col2" colwidth="*" align="right"/>
+       <colspec colnum="3" colname="col3" colwidth="*" align="right"/>
+       <thead>
+         <row>
+           <entry>Feature</entry>
+           <entry>Bit</entry>
+           <entry>Value</entry>
+         </row>
+       </thead>
+       <tbody>
+         <row>
+           <entry>UID</entry>
+           <entry>0</entry>
+           <entry>0x1</entry>
+         </row>
+         <row>
+           <entry>NOSRCADDR</entry>
+           <entry>1</entry>
+           <entry>0x2</entry>
+         </row>
+         <row>
+           <entry>FLOCK</entry>
+           <entry>3</entry>
+           <entry>0x8</entry>
+         </row>
+         <row>
+           <entry>SUBSCRIBE2</entry>
+           <entry>4</entry>
+           <entry>0x10</entry>
+         </row>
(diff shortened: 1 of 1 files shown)


xml/bp_troubleshooting_cephfs.xml
```diff
@@ -271,4 +271,382 @@ sudo zypper in ceph-common
...
+       </tbody>
+     </tgroup>
+   </table>
+   <para>
+     A possible solution to allow old kernel clients to mount a recent &cephfs;
+     is to modify the CRUSH profile.  For example, to fix the problem shown
+     above, the following command can be used:
+   </para>
+<screen>
+ceph osd crush tunables hammer
```
asettle, line 639:
```
&prompt.cephuser;ceph osd crush tunables hammer

File: xml/bp_troubleshooting_cephfs.xml [hunk 2]
```diff
@@ -271,4 +271,382 @@ sudo zypper in ceph-common
...
+</screen>
+    <para>
+      This will adjust the CRUSH profile to the behaviour it had for the Hammer
+      (0.94) release.  Note however that this is not the optimal behaviour for
+      the cluster.  To change back to the optimal profile, run the following
+      command:
+    </para>
+<screen>
+ceph osd crush tunables optimal
```
asettle, line 648:
``` suggestion
&prompt.cephuser;ceph osd crush tunables optimal

**Comment by luis-henrix** [comment]
By the way: I haven't rendered the document, so I'm not sure how it looks like.  And maybe that table is overkill... if it is, I'm OK dropping it (don't worry, it's automatically generated ;-) ).

**Comment by asettle** [comment]
> By the way: I haven't rendered the document, so I'm not sure how it looks like. And maybe that table is overkill... if it is, I'm OK dropping it (don't worry, it's automatically generated ;-) ).

Here's a little snippet :) 

Screenshot from 2020-09-29 15-12-45

File: xml/bp_troubleshooting_cephfs.xml [hunk 3]
```diff
@@ -271,4 +271,382 @@ sudo zypper in ceph-common
      </para>
    </sect2>
  </sect1>
+ <sect1 xml:id="bp-troubleshooting-old-kernel-clients">
+   <title>Using legacy kernel clients</title>
```
lproven, line 275: I think we should explain very briefly what a "kernel client" is. (I did not know.)

1 paragraph would be enough.

xml/bp_troubleshooting_cephfs.xml
```diff
@@ -271,4 +271,382 @@ sudo zypper in ceph-common
...
+   <para>
+     Sometimes it is necessary to connect to a &cephfs; instance using old
+     kernel clients that may not support all the features required by the
+     cluster.  When this happens, the kernel client will fail to mount the
+     filesystem and will emit messages similar to the one shown below:
```
lproven, line 280: This is ambiguous.

As I understand it, from a quick Google and bit of background reading, there is a Ceph client built into the Linux kernel. That should be specified and explained, at least briefly.

But here this talks about the _version_ of the kernel client.

That is confusing.

I presume that "client" here in general means "machines which are clients to the Ceph cluster." Is that correct?

I ask because I see 2 possible interpretations of "client" here: the Ceph client, versus the client computer.

So when we talk about versions, do we mean:

[1] The version of the Ceph client built into the kernel running on client machines?

(Is it possible that specific Linux distro/versions can have different versions of the Ceph client software? Because that is what this interpretation implies.)

[2] The version of the Linux kernel on the client machine(s)? 

(That is implied if the Ceph client version is tied to the version of the kernel.)

[3] The version of the Linux distribution running on client machines?

(Generally, assuming that the client machines' OS is up to date -- itself a **big** assumption -- then it seems to me that a specific version of the client OS will have a specific fixed version of the kernel, and if so, then to upgrade the Ceph client on the client machine means upgrading the kernel which means upgrading the distro version.)

xml/bp_troubleshooting_cephfs.xml
```diff
@@ -271,4 +271,382 @@ sudo zypper in ceph-common
...
+     A possible solution to allow old kernel clients to mount a recent &cephfs;
+     is to modify the CRUSH profile.  For example, to fix the problem shown
+     above, the following command can be used:
+   </para>
+<screen>
+&prompt.cephuser;ceph osd crush tunables hammer
+</screen>
+    <para>
+      This will adjust the CRUSH profile to the behaviour it had for the Hammer
+      (0.94) release.  Note however that this is not the optimal behaviour for
```
lproven, line 643: If this value it a text string, then without considerable prior knowledge, it is not possible to work out what values can be used. 

Therefore, I think we should specify a list of permissible values and what they mean. I think that probably means another table, I am afraid.

I would suggest at a minimum, the following values:
col1: name of release (e.g. "hammer")
col2: version number or date of release (or some other value allowing them to be ordered)
col3: SES version that equates to that release

The columns need not be in that order, but I don't think we can just leave people to try to Google what word they need to insert here to turn certain features on or off.

lproven (changes requested):

xml/bp_troubleshooting_cephfs.xml
luis-henrix: Right, I'll try to rephrase the initial paragraphs to make it a bit more clear.  However, my opinion is that a troubleshooting guide necessarily needs to assume some prior knowledge -- defining what a "kernel client" is could be something for another guide.

(Thanks a lot for your review, BTW!)

xml/bp_troubleshooting_cephfs.xml
luis-henrix: Right, I'm rephrasing the whole thing trying to clarify this.

xml/bp_troubleshooting_cephfs.xml
luis-henrix: Ok, makes sense.  I'll add a table or a list for this.  However, I don't think the corresponding SES version  is required because AFAIK the first release we support CephFS (SES5) is more recent than any value accepted for the profiles.  Let's see what I can come up with.

xml/bp_troubleshooting_cephfs.xml
lproven: > Right, I'll try to rephrase the initial paragraphs to make it a bit more clear. However, my opinion is that a troubleshooting guide necessarily needs to assume some prior knowledge -- defining what a "kernel client" is could be something for another guide.

I take your point, yes -- although we could afford a sentence. :-) The question is, though, whether the kernel client is tied to the kernel version, and if the kernel version is tied to the client machine's OS version.

(For example, in Ubuntu releases "hardware enablement stacks" for its Long Term Support (LTS) releases. These contain a kernel, X server etc. backported from the current short-term support release. This means that it is possible to stay with a given OS release (say, 18.04) and nonetheless install a newer kernel. 

> (Thanks a lot for your review, BTW!)

You're welcome. It's what I am here for. :-)

xml/bp_troubleshooting_cephfs.xml
lproven: OK, fair, but it might be useful reference for customers who have not memorised a list of Ceph releases. Many people are too busy for things like this.

In case it is helpful:
https://docs.ceph.com/en/latest/releases/
https://docs.ceph.com/en/latest/releases/archived-index/
I think we don't need all the point-releases -- just those where the name changes, at the most.

asettle: In fairness, we _assume_ the customer is familiar with Ceph and their release names. I don't see this as an issue.

luis-henrix:
I've just pushed another revision of this new section.  Hopefully I've addressed all the change requests (and the removal of double-space between sentences!).

xml/bp_troubleshooting_cephfs.xml
```diff
@@ -271,4 +271,450 @@ sudo zypper in ceph-common
      </para>
    </sect2>
  </sect1>
+ <sect1 xml:id="bp-troubleshooting-old-clients-mount">
+   <title>Mounting CephFS Using Old Kernel Clients</title>
+   <para>
+     The Linux kernel available in the &sls; 15 SP2 Operating System includes a
+     &cephfs; client that is able to take full advantage of all the features
+     available on an SES7 cluster. All relevant features and bug fixes are
+     backported to this Operating System.
```
lproven, line 280:
```suggestion
     The kernel in &sls; 15 SP2 includes a
     &cephfs; client that is able to take full advantage of all the features
     available on an SES7 cluster. All relevant features and bug fixes are
     backported to this operating system.
```

lproven (changes requested):
This is much better -- I've just shortened it slightly. 

Great work -- thank you for this contribution!

xml/bp_troubleshooting_cephfs.xml
```diff
@@ -271,4 +271,450 @@ sudo zypper in ceph-common
      </para>
    </sect2>
  </sect1>
+ <sect1 xml:id="bp-troubleshooting-old-clients-mount">
+   <title>Mounting CephFS Using Old Kernel Clients</title>
```
asettle, line 275:
```suggestion
   <title>Mounting &cephfs; Using Old Kernel Clients</title>

File: xml/bp_troubleshooting_cephfs.xml [hunk 8]
```diff
@@ -271,4 +271,450 @@ sudo zypper in ceph-common
...
+           <entry>62</entry>
+           <entry>0x4000000000000000</entry>
+         </row>
+       </tbody>
+     </tgroup>
+   </table>
+   <para>
+     A possible solution to allow an old kernel client to mount a recent
+     &cephfs; is to modify the cluster CRUSH profile. CRUSH profiles define a
+     set of CRUSH tunables that are named after the Ceph versions in which they
```
asettle, line 643:
```suggestion
     set of CRUSH tunables that are named after the &ceph; versions in which they

**Review by asettle — APPROVED** [review]
Tiny nits, but looks good!

**Comment by luis-henrix** [comment]
> Tiny nits, but looks good!

Thanks, all suggestions committed.

**Comment by luis-henrix** [comment]
Sorry about my github illiteracy, but I'm still seeing "1 review requesting changes" although I think I've addressed all the requests.  Or have I missed some comment?

**Comment by asettle** [comment]
@lproven let me know when you have reviewed again, as we need to move forward on this one :) thanks!

File: xml/bp_troubleshooting_cephfs.xml [hunk 9]
```diff
@@ -271,4 +271,450 @@ sudo zypper in ceph-common
...
+ <sect1 xml:id="bp-troubleshooting-old-clients-mount">
+   <title>Mounting &cephfs; Using Old Kernel Clients</title>
+   <para>
+     The kernel in &sls; 15 SP2 includes a
+     &cephfs; client that is able to take full advantage of all the features
+     available on an SES7 cluster. All relevant features and bug fixes are
+     backported to this operating system.
+   </para>
+   <para>
+     However, it may be necessary to access &cephfs; from another systems that
```
lproven, line 283:
```suggestion
     However, it may be necessary to access &cephfs; from other systems that
```

xml/bp_troubleshooting_cephfs.xml
```diff
@@ -271,4 +271,450 @@ sudo zypper in ceph-common
...
+     may provide an older &cephfs; client, which may not support all the
+     features required by an SES7 cluster. When this happens, the kernel client
```
lproven, line 285:
```suggestion
     features required by an &productname; &productnumber; cluster. When this happens, the kernel client
```

xml/bp_troubleshooting_cephfs.xml
```diff
@@ -271,4 +271,450 @@ sudo zypper in ceph-common
...
+           <entry>0x4000000000000000</entry>
+         </row>
+       </tbody>
+     </tgroup>
+   </table>
+   <para>
+     A possible solution to allow an old kernel client to mount a recent
+     &cephfs; is to modify the cluster CRUSH profile. CRUSH profiles define a
+     set of CRUSH tunables that are named after the &ceph; versions in which they
+     were introduced. For example, the 'firefly' tunables are first supported in
```
lproven, line 644:
```suggestion
     were introduced. For example, the <literal>firefly</literal> tunables are first supported in
```

xml/bp_troubleshooting_cephfs.xml
```diff
@@ -271,4 +271,450 @@ sudo zypper in ceph-common
...
+     the Firefly release (0.80) and older clients will not be able to access the
```
lproven, line 645:
```suggestion
     the Firefly release (0.80), and older clients will not be able to access the
```

xml/bp_troubleshooting_cephfs.xml
```diff
@@ -271,4 +271,450 @@ sudo zypper in ceph-common
...
+</screen>
+   <para>
+     The following table lists the available CRUSH profiles and which CRUSH
+     tunables versions (the CRUSH_TUNABLE feature bits in the previous table)
+     they correspond to. It also identifies the minimum kernel version required
+     to use for each profile. Note however that Operating System vendors may
+     choose to backport features to their kernels, so these kernel versions are
+     valid for mainline kernels only. The kernel client included in &sls; 15
+     SP2, for example, includes backports of features and bug fixes relevant for
+     usage in SES7 clusters.
```
lproven, line 668: @asettle -- what do you think, entities here too or leave the short form?

lproven (changes requested):

xml/bp_troubleshooting_cephfs.xml
asettle: Hmm, let's leave short form for now :)

PR #617, opened 2020-09-29
Adding igw install instructions

asettle:
Also need to further update the documentation to reflect previous iterations: https://susedoc.github.io/doc-ses/ses6/single-html/ses-deployment/#cha-ceph-as-iscsi

### DIFF
--- a/xml/deploy_cephadm.xml
+++ b/xml/deploy_cephadm.xml
@@ -1415,6 +1415,28 @@ placement:
   - ses-min3
 </screen>
    </sect3>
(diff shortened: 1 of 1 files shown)


mgfritch (approved):
it's also possible to optionally pass a `trusted_ip_list` list, but otherwise lgtm!

asettle:
> it's also possible to optionally pass a `trusted_ip_list` list, but otherwise lgtm!

I've added to the bug report so we can track the work there :)

PR #622, opened 2020-10-01
Renaming "upgrade" to "update" to reflect proper terms

asettle:


jan--f (approved):
LGTM. Wondering if we should explicitly point out the disparity in naming? We call it update, but the command is upgrade. Follow-up PR would be better though.

PR #632, opened 2020-10-02
Removing ceph.conf for RGW

asettle:


xml/admin_ceph_gateway.xml
```diff
@@ -3001,24 +3001,6 @@ rgw_override_bucket_index_max_shards = 12
...
-<screen>[client.rgw.<replaceable>INSTANCE-NAME</replaceable>]
-...
-rgw_zone=<replaceable>ZONE-NAME</replaceable></screen>
-    <para>
-     For example:
-    </para>
-<screen>[client.rgw.rgw2]
-host = rgw2
-rgw frontends = "civetweb port=80"
-rgw_zone=us-west</screen>
```
theanalyst, line 3021: just replace this conf section with `ceph config set <service-name> rgw_zone us-west `...

xml/admin_ceph_gateway.xml
```diff
@@ -1927,7 +1927,7 @@ number_of_objects_expected_in_a_bucket / 100000
    <para>
     Be aware that the maximum number of shards is 7877.
    </para>
-   <sect3>
+<!--   <sect3 xml:id="rgw-simple-config">
```
asettle, line 1930: @theanalyst - what about this section? Remove?

theanalyst: I don't think we need to remove as such. The params themselves are the same, just the delivery is different. just that there are 2 ways of configuring RGW, using `ceph config` and of course updating the ceph.conf files, though former is more preferred

theanalyst (approved):

PR #634, opened 2020-10-02
SES7 Deployment Guide/Hardware Recommendations: Updating networking command

asettle:


lproven (approved):

xml/deployment_hwrecommend.xml
```diff
@@ -173,9 +173,13 @@
...
      in different rooms and served by different switches, you need to specify
-     their public network address:
+     their public network address in CIDR notation:
     </para>
-<screen>&prompt.cephuser;ceph config set mon public_network <replaceable>MON_NETWORK</replaceable></screen>
+<screen>&prompt.cephuser;ceph config set mon public_network "<replaceable>MON_NETWORK_1</replaceable>, <replaceable>MON_NETWORK_2</replaceable>, <replaceable>MON_NETWORK_N</replaceable></screen>
+    <para>
+      For example:
+    </para>
+<screen>&prompt.cephuser;ceph config set mon public_network "1.3.4.0/24, 1.3.43.0/24, 1.3.42.0/24"</screen>
```
smithfarm, line 182: Please change `"1.3.4.0/24, 1.3.43.0/24, 1.3.42.0/24"` to:

```
"192.168.1.0/24, 10.10.0.0/16"
```

lproven: I considered that, but we've had this discussion before.

There's also the 172.16.0.0–172.31.255.255 range, of course.

smithfarm: @lproven We cannot use `1.3.4.0/24, 1.3.43.0/24, 1.3.42.0/24` because it would look like we are plagiarizing https://access.redhat.com/solutions/1463363

;-)

smithfarm: @lproven Of course, if you'd like to add `172.16.0.0–172.31.255.255` as well, please provide the CIDR equivalent.

xml/deployment_hwrecommend.xml
```diff
@@ -173,9 +173,13 @@
...
    </sect3>
   </sect2>
  </sect1>
```
smithfarm, line 185: At the bottom of this section, please add a WARNING box:

```
IMPORTANT: If you do specify more than one network segment for the public (or cluster) network as described in this section, each of these subnets must be capable of routing to all the others - otherwise, the MONs and other Ceph daemons on different network segments will not be able to communicate and a split cluster will ensue. Additionally, if you are using a firewall, make sure you include each IP address/subnet in your IP tables and open ports for them on all nodes as necessary.
```

smithfarm (approved):

PR #640, opened 2020-10-05
Fixed CephFS client mount command line (#639)

### LINKED ISSUE #639: SES7 Admin Guide: CephFS: 13.1.1 Client Preparation
### Type of report
   - [x] bug
   - [ ] feature request

### Source Document
13.1.1  Client Preparation:

https://susedoc.github.io/doc-ses/master/single-html/ses-admin/#cephfs-client-preparation

### Summary

At the bottom of this section there is an "! Important" box which says:

    Without the ceph-common package (and thus without the mount.ceph helper), 
    the monitors' IPs will need to be used instead of their names. This is because 
    the kernel client will be unable to perform name resolution.

which is fine. But then it says:

    The basic mount syntax is:

    cephadm@adm > sudo mount -t ceph monip[:port][,monip2[:port]...]:/[subdir] mnt

This "basic mount syntax" is not complete, and I don't think the "sudo" is appropriate. Every Linux user proficient with the command line knows that "mount" requires root privileges. So, I suggest to change the mount command to:

    # mount -t ceph monip[:port][,monip2[:port]...]:/[subdir] mnt -o name=[cephx user name],secret=[secret string]

tbazant:
only the XML file change is relevant, the images are part of this only because they were not optimized (decreased in size) by `optipng`

Rendered version: 
obrazek

closes #639


smithfarm:
Asking @jan--f to approve the text change (as @tbazant notes in PR description, we don't have to consider the image file changes)

smithfarm:
(Note: this change is based on tests I did yesterday.)

asettle (approved):

xml/admin_ceph_cephfs.xml
```diff
@@ -56,7 +56,8 @@
        The basic mount syntax is:
      </para>
 <screen>
-&prompt.cephuser;sudo mount -t ceph monip[:port][,monip2[:port]...]:/[subdir] mnt
+&prompt.root;mount -t ceph <replaceable>MON1_IP</replaceable>[:<replaceable>PORT</replaceable>],<replaceable>MON2_IP</replaceable>[:<replaceable>PORT</replaceable>],...:/<replaceable>OPTIONAL_SUBDIR</replaceable> \
```
jan--f, line 59:
```suggestion
&prompt.root;mount -t ceph <replaceable>MON1_IP</replaceable>[:<replaceable>PORT</replaceable>],<replaceable>MON2_IP</replaceable>[:<replaceable>PORT</replaceable>],...:<replaceable>CEPHFS_MOUNT_TARGET</replaceable> \
```
Lets make it clear that a directory is needed after the last `:`. In most cases this `/` but I think this is more straight forward then saying you need `/` and an optional subdir (but without the `/`).

jan--f (approved):
I think this is ok. One proposal for a nit picky change though.

xml/admin_ceph_cephfs.xml
smithfarm: good point :+1:

Issue #662, opened 2020-10-07
SES7 Deployment Guide 5.3.4  Deploying Services and Gateways


smithfarm:
### Type of report
   - [x] bug
   - [ ] feature request

### Source Document
5.3.4  Deploying Services and Gateways:

https://susedoc.github.io/doc-ses/master/single-html/ses-deployment/#day2-deployment

5.4 Deploying Services and Gateways:

https://susedoc.github.io/doc-ses/master/single-html/ses-deployment/#deploy-cephadm-day2

### Summary

1. The whole section "5.3.4  Deploying Services and Gateways" is bogus and should be removed

2. "5.4 Deploying Services and Gateways" is missing an intro, which can be taken from the bogus section "5.3.4  Deploying Services and Gateways"

Issue #661, opened 2020-10-07
SES7 Deployment Guide: 5.3.3  Update Nodes and Deploy Cluster


smithfarm:
### Type of report
   - [x] bug
   - [ ] feature request

### Source Document
5.3.3  Update Nodes and Deploy Cluster:

https://susedoc.github.io/doc-ses/master/single-html/ses-deployment/#deploy-cephadm-deploy

### Summary

The title of this section should be

    Update Nodes and Bootstrap Cluster

And the section should educate the user that `ceph-salt apply` runs `cephadm bootstrap` to bootstrap the cluster. When bootstrapping is complete, the cluster will have 1 MON and 1 MGR. Nothing else. This is also sometimes referred to as "Day 1 Deployment".

After bootstrapping is complete, the user should continue with "Day 2 Deployment" using service spec YAML files and `ceph orch apply`. This is covered in the subsequent chapters.

smithfarm:
Instead of "Day 1" and "Day 2" (which are developer terms and might be confusing for users) you could refer to the "bootstrapping phase" and "post-bootstrap deployment phase".

In short, deploying a SES7 cluster is a two-phase process:

Phase One (bootstrapping) is `ceph-salt apply`

Phase Two (post-bootstrap) is `ceph orch apply`.

Issue #659, opened 2020-10-06
[doc] 2.3.2  Recommended Production Cluster Configuration


swiftgist:
### Type of report
   - [ ] bug
   - [ ] feature request

### Source Document
2.3.2  Recommended Production Cluster Configuration:

https://susedoc.github.io/doc-ses/master/single-html/ses-deployment/#ses-bp-production-cluster

### Summary
Missing any discussion of setting the public network

### Expected Results
This may require another opinion.  One argument is that SES5 and SES6 *always* set the public and cluster network.  The current default behavior with SES7 is that neither is set.  This documentation gives the `cluster network` command but does say it's optional.

### Actual Results
When a ceph cluster has neither the public nor cluster network set, the ceph-osd process listens on all interfaces.

```
# ss -nltp
State      Recv-Q     Send-Q           Local Address:Port           Peer Address:Port                                                                                               
LISTEN     0          128                    0.0.0.0:6801                0.0.0.0:*         users:(("ceph-osd",pid=22974,fd=16))                                                     
LISTEN     0          128                    0.0.0.0:6802                0.0.0.0:*         users:(("ceph-osd",pid=22974,fd=17))                                                     
LISTEN     0          128                    0.0.0.0:6803                0.0.0.0:*         users:(("ceph-osd",pid=22974,fd=18))                                                     
LISTEN     0          128                    0.0.0.0:6804                0.0.0.0:*         users:(("ceph-osd",pid=22974,fd=19))                                                     
LISTEN     0          128                    0.0.0.0:6805                0.0.0.0:*         users:(("ceph-osd",pid=22974,fd=20))                                                     
LISTEN     0          128                    0.0.0.0:6806                0.0.0.0:*         users:(("ceph-osd",pid=22974,fd=21))                                                     
LISTEN     0          128                    0.0.0.0:22                  0.0.0.0:*         users:(("sshd",pid=1143,fd=3))                                                           
LISTEN     0          128                    0.0.0.0:6807                0.0.0.0:*         users:(("ceph-osd",pid=22974,fd=22))                                                   
```

In environments with only a single network interface, I do not see an issue.  In environments with multiple interfaces dedicated for specific purposes such as a management network separated from the intended public and cluster network, I can imagine clients connecting over the _other_ interfaces.

### Notes
If the public and cluster network are both set, then ceph-osd listens only on the desired interfaces and matches the behavior for SES5 and SES6.

```
# ss -nltp
[...]

PR #664, opened 2020-10-07
Editing deployment file

### LINKED ISSUE #663: SES7 Deployment Guide: 5.3  Deploying the Ceph Cluster
### Type of report
   - [x] bug
   - [ ] feature request

### Source Document
5.3  Deploying the Ceph Cluster:

https://susedoc.github.io/doc-ses/master/single-html/ses-deployment/#deploy-cephadm-day1

### Summary

This chapter is missing a "Final Steps" section. I recommend to put this section at the very end of the chapter, with the following content:

5.3.999 Final Steps

Once `ceph-salt apply` has completed successfully, you should have 1 MON and 1 MGR running on the cluster's bootstrap minion, and you should be able to run `ceph status` successfully on any of the minions that were given the `admin` role, either as `root` or the user `cephadm`.

Congratulations! You have completed the bootstrapping phase (also sometimes referred to as "Day 1 deployment"). The next steps involve using the Cephadm Orchestrator to deploy additional MONs, MGRs, OSDs, the Monitoring Stack, and Gateways. These steps are collectively referred to as the "post-bootstrapping phase", or sometimes "Day 2 deployment".

Before you embark on that adventure, though, now would be a good time to review your new cluster's network settings. At this point, the `public_network` setting has been populated based on what you entered for `/cephadm_bootstrap/mon_ip` in the ceph-salt configuration, but this setting applies only to the MONs. You can review this setting by doing:

    ceph config get mon public_network

This is the minimum that Ceph requires to work, but it is recommended (and you may find it useful) to make this `public_network` setting "global", which means it will apply to all types of Ceph daemons, and not just to MONs:

    ceph config set global public_network "$(ceph config get mon public_network)"

This is not required, but if you don't do it, the Ceph OSDs and other daemons (except MONs) will listen on **all addresses**, which may not be what you want.

If you want your OSDs to communicate amongst themselves using a completely separate network, now is a good time to set that up:

    ceph config set global cluster_network "<cluster_network_in_cidr_notation>"

That way, the OSDs you create during "Day 2 deployment" will use the intended cluster network right from the start.

asettle:
1. Update title of minimum config chap
2. Add final steps section (extra for networking)
3. Added note about the ceph-salt config command
4. Updated the ceph prompts for salt

Closes #663, #662, #661, #659

### DIFF
--- a/xml/deploy_cephadm.xml
+++ b/xml/deploy_cephadm.xml
@@ -295,6 +295,13 @@ minion1:
     Use the <command>ceph-salt config</command> command to configure the basic
     properties of the cluster.
    </para>
+   <important>
+     <para>
+       The <filename>/etc/ceph/ceph.conf</filename> file is managed by &cephadm;
+       and users <emphasis>should not</emphasis> edit it. &ceph; configuration
+       parameters should be set using the new <command>ceph config</command> command.
+     </para>
+   </important>
    <sect3 xml:id="deploy-cephadm-configure-shell">
     <title>&cephsalt; Shell</title>
     <para>
@@ -1018,7 +1025,7 @@ config: OK
   </sect2>
 
   <sect2 xml:id="deploy-cephadm-deploy">
-   <title>Update Nodes and Deploy Cluster</title>
+   <title>Update Nodes and Bootstrap Minimal Cluster</title>
    <para>
     Before you deploy the cluster, update all software packages on all nodes:
    </para>
@@ -1034,7 +1041,7 @@ config: OK
    </para>
 <screen>&prompt.smaster;ceph-salt update --reboot</screen>
    <para>
-    or reboot them in a separate step:
+    Or, reboot them in a separate step:
    </para>
 <screen>&prompt.smaster;ceph-salt reboot</screen>
    <important>
@@ -1045,15 +1052,20 @@ config: OK
     </para>
    </important>
    <para>
-    After the nodes are updated, deploy the minimal cluster:
+    After the nodes are updated, bootstrap the minimal cluster:
    </para>
 <screen>&prompt.smaster;ceph-salt apply</screen>
+   <note>
+     <para>
+       When bootstrapping is complete, the cluster will have one &mon; and one &mgr;.
+     </para>
+   </note>
    <para>
     The above command will open an interactive user interface that shows the
     current progress of each minion.
    </para>
    <figure>
-    <title>Deployment of Minimal Cluster</title>
+    <title>Deployment of a Minimal Cluster</title>
     <mediaobject>
      <imageobject role="fo">
       <imagedata fileref="cephadm_deploy.png" width="75%"/>
@@ -1074,23 +1086,65 @@ config: OK
 <screen>&prompt.smaster;ceph-salt apply --non-interactive</screen>
    </tip>
   </sect2>
-
-  <sect2 xml:id="day2-deployment">
-   <title>Deploying Services and Gateways</title>
-   <para>
-    After you have deployed the basic &ceph; cluster, you need to deploy core
-    services to more cluster nodes. To make the cluster data accessible to
-    clients, deploy additional services as well.
-   </para>
-   <para>
-    Currently, we support deployment of &ceph; services on the command line by
-    using the &ceph; orchestrator (<command>ceph orch</command> subcommands).
-   </para>
+  <sect2 xml:id="deploy-min-cluster-final-steps">
+    <title>Final Steps</title>
+    <para>
+      Once the <command>ceph-salt apply</command> command has completed,
+      you should have one MON and one MGR running on the
+      cluster's bootstrap minion. You should be able to run
+      the <command>ceph status</command> command successfully on any of the
+      minions that were given the admin role, either as <literal>root</literal>
+      or the <literal>cephadm</literal> user.
+    </para>
+    <para>
+      The next steps involve using the &cephadm; to deploy additional &mon;,
+      &mgr;, OSDs, the Monitoring Stack, and Gateways. These steps are
+      collectively referred to as the "post-bootstrapping phase", or sometimes
+      "Day 2 deployment".
+    </para>
+    <para>
+      Before you continue, review your new cluster's network settings. At this
+      point, the <literal>public_network</literal> setting has been populated
+      based on what was entered for <literal>/cephadm_bootstrap/mon_ip</literal>
+      in the <literal>ceph-salt</literal> configuration. However, this setting
+      was only applied to &mon;. You can review this setting with the following
+      command:
+    </para>
+<screen>&prompt.smaster;ceph config get mon public_network</screen>
+    <para>
+      This is the minimum that &ceph; requires to work, but we recommend
+      making this <literal>public_network</literal> setting <literal>global</literal>,
+      which means it will apply to all types of &ceph; daemons, and not just to MONs:
+    </para>
+<screen>&prompt.smaster;ceph config set global public_network "$(ceph config get mon public_network)"</screen>
+    <note>
+      <para>
+        This step is not required. However, if you do not use this setting, the
+        &ceph; OSDs and other daemons (except &mon;) will listen on <emphasis>all addresses</emphasis>.
+      </para>
+      <para>
+        If you want your OSDs to communicate amongst themselves using a
+        completely separate network, run the following command:
+      </para>
+<screen>&prompt.smaster;ceph config set global cluster_network "<replaceable>cluster_network_in_cidr_notation</replaceable>"</screen>
+      <para>
+        Executing this command will ensure that the OSDs created in your
+        deployment will use the intended cluster network from the start.
+      </para>
+    </note>
   </sect2>
- </sect1>
-  
+
  <sect1 xml:id="deploy-cephadm-day2">
   <title>Deploying Services and Gateways</title>
+  <para>
+   After you have deployed the basic &ceph; cluster, you need to deploy core
+   services to more cluster nodes. To make the cluster data accessible to
+   clients, deploy additional services as well.
+  </para>
+  <para>
+   Currently, we support deployment of &ceph; services on the command line by
+   using the &ceph; orchestrator (<command>ceph orch</command> subcommands).
+  </para>
 
   <sect2 xml:id="deploy-cephadm-day2-orch">
    <title>The <command>ceph orch</command> Command</title>
--- a/xml/deploy_cephadm.xml
+++ b/xml/deploy_cephadm.xml
@@ -300,6 +300,7 @@ minion1:
        The <filename>/etc/ceph/ceph.conf</filename> file is managed by &cephadm;
        and users <emphasis>should not</emphasis> edit it. &ceph; configuration
        parameters should be set using the new <command>ceph config</command> command.
+       See <xref linkend="cha-ceph-configuration-db"/> for more information.
      </para>
    </important>
    <sect3 xml:id="deploy-cephadm-configure-shell">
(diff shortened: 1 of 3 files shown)


xml/deploy_cephadm.xml
```diff
@@ -295,6 +295,13 @@ minion1:
     Use the <command>ceph-salt config</command> command to configure the basic
     properties of the cluster.
    </para>
+   <important>
+     <para>
+       The <filename>/etc/ceph/ceph.conf</filename> file is managed by &cephadm;
+       and users <emphasis>should not</emphasis> edit it. &ceph; configuration
+       parameters should be set using the new <command>ceph-salt config</command> command.
```
theanalyst, line 302: just `ceph config` here and not `ceph-salt config`

xml/deploy_cephadm.xml
```diff
@@ -295,6 +295,13 @@ minion1:
     Use the <command>ceph-salt config</command> command to configure the basic
     properties of the cluster.
    </para>
+   <important>
+     <para>
+       The <filename>/etc/ceph/ceph.conf</filename> file is managed by &cephadm;
+       and users <emphasis>should not</emphasis> edit it. &ceph; configuration
+       parameters should be set using the new <command>ceph config</command> command.
```
tbazant, line 302: maybe add ref to #cha-ceph-configuration-db ?

tbazant (approved):
small hint only

xml/deploy_cephadm.xml
```diff
@@ -1074,23 +1086,65 @@ config: OK
...
-   </para>
-   <para>
-    Currently, we support deployment of &ceph; services on the command line by
-    using the &ceph; orchestrator (<command>ceph orch</command> subcommands).
-   </para>
+  <sect2 xml:id="deploy-min-cluster-final-steps">
+    <title>Final Steps</title>
+    <para>
+      Once the <command>ceph-salt apply</command> command has completed,
+      you should have one MON and one MGR running on the
```
lproven, line 1094: Is there a particular reason that sometimes we use an entity for MON and sometimes not?

lproven (changes requested):

xml/deploy_cephadm.xml
asettle: Forgetfulness? But for the same reason why we'd use SES vs. &productname; - the shorthand.

xml/deploy_cephadm.xml
```diff
@@ -1074,23 +1086,65 @@ config: OK
...
+      cluster's bootstrap minion. You should be able to run
+      the <command>ceph status</command> command successfully on any of the
+      minions that were given the admin role, either as <literal>root</literal>
```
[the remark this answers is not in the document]
asettle: Ah, yes, we should. That's a miss on my part.

xml/deploy_cephadm.xml
```diff
@@ -24,6 +24,17 @@
...
+   <para>
+     It is important to note that the &ceph; community documentation uses the
+     <command>cephadm bootstrap</command> command during initial deployment.
+     When using &productname; &productnumber;, the &cephsalt; tool does this for
+     you and you should not run the <command>cephadm bootstrap</command>
+     command at any point during your deployment. Any &ceph; cluster deployment
+     manually using the <command>cephadm bootstrap</command> will be
+     unsupported.
+   </para>
+ </important>
```
swiftgist, line 37: I"m being nitpicky, so use your discretion. Remove the passive voice and second person.

---
Note that the &ceph; community documentation uses the <command>**cephadm bootstrap**</command> during initial deployment.  The &cephsalt; tool calls the <command>**cephadm bootstrap**</command> and should not be run directly.  Any &ceph; cluster deployment manually using the <command>**cephadm bootstrap**</command> will be unsupported.

xml/deploy_cephadm.xml
```diff
@@ -1074,23 +1099,66 @@ config: OK
...
-   <para>
-    Currently, we support deployment of &ceph; services on the command line by
-    using the &ceph; orchestrator (<command>ceph orch</command> subcommands).
-   </para>
+  <sect2 xml:id="deploy-min-cluster-final-steps">
+    <title>Final Steps</title>
+    <para>
+      Once the <command>ceph-salt apply</command> command has completed,
+      you should have one &mon; and one &mgr; running on the
+      cluster's bootstrap minion. You should be able to run
```
swiftgist, line 1107: The bootstrap minion will be running one &mon; and one &mgr;.

xml/deploy_cephadm.xml
```diff
@@ -1074,23 +1099,66 @@ config: OK
...
+      the <command>ceph status</command> command successfully on any of the
+      minions that were given the <literal>admin</literal> role, either as <literal>root</literal>
+      or the <literal>cephadm</literal> user.
+    </para>
+    <para>
+      The next steps involve using the &cephadm; to deploy additional &mon;,
+      &mgr;, OSDs, the Monitoring Stack, and Gateways. These steps are
+      collectively referred to as the "post-bootstrapping phase", or sometimes
+      "Day 2 deployment".
```
swiftgist, line 1116: Day 2 is operations.  At least according to all the sources that I can find.  Day 1 is deployment.  I don't know that having only a single monitor and manager is considered a successful deployment from a customer perspective.

xml/deploy_cephadm.xml
```diff
@@ -1074,23 +1099,66 @@ config: OK
...
+    <note>
+      <para>
+        This step is not required. However, if you do not use this setting, the
+        &ceph; OSDs and other daemons (except &mon;) will listen on <emphasis>all addresses</emphasis>.
+      </para>
+      <para>
+        If you want your OSDs to communicate amongst themselves using a
+        completely separate network, run the following command:
+      </para>
+<screen>&prompt.smaster;ceph config set global cluster_network "<replaceable>cluster_network_in_cidr_notation</replaceable>"</screen>
```
swiftgist, line 1142: Wouldn't this be `ceph config set osd cluster_network` like https://susedoc.github.io/doc-ses/master/single-html/ses-deployment/#storage-bp-net-private?

xml/deploy_cephadm.xml
```diff
@@ -1074,23 +1099,66 @@ config: OK
...
+    </note>
   </sect2>
- </sect1>
-  
+</sect1>
+
  <sect1 xml:id="deploy-cephadm-day2">
   <title>Deploying Services and Gateways</title>
+  <para>
+   After you have deployed the basic &ceph; cluster, you need to deploy core
```
swiftgist, line 1154: After deploying the basic &ceph; cluster, deploy core

smithfarm (approved):

xml/deploy_cephadm.xml
smithfarm: Upstream recommends `global`. I'm pretty sure both `public_network` and `cluster_network` should be set globally, not just for a particular type of daemon.

https://docs.ceph.com/en/latest/rados/configuration/network-config-ref/#cluster-network

Cluster Network

If you declare a cluster network, OSDs will route heartbeat, object replication and recovery traffic over the cluster network. This may improve performance compared to using a single network. To configure a cluster network, add the following option to the [global] section of your Ceph configuration file.

```
[global]
        # ... elided configuration
        cluster network = {cluster-network/netmask}
```

xml/deploy_cephadm.xml
smithfarm: The `ceph orch apply` commands for initially deploying services and daemons are identical to the commands used in day-to-day operations. I guess that's why the orchestrator folks refer to this deployment phase as "Day 2".

Be that as it may, I'm fine to drop all mention of days and numbers thereof, in order to avoid conclusion. "Bootstrap" and "post-bootstrap" deployment phases are plenty fine for our terminological needs, and hopefully impossible to misconstrue.

swiftgist:
Aft reading the section on the cluster network settings, I feel like there's a blue dog in the room that nobody wants to talk about.  The difference between global, mon, osd and other components for both public_network and cluster_network.

Is this one of these cases where we have useless knobs?  To my knowledge, only the storage nodes use the cluster network.  None of the gateways, monitors or managers require or would even use the cluster network.  

On the public side, I'm assuming (which gets anyone in trouble) that the purpose of a global vs. component setting is that the component setting overrides a global setting if there is one.  The end result is several possibilities

- None are set and all Ceph components bind to all network interfaces
- The global public_network is set and all Ceph components to bind to the public interfaces
- The global public_network is *not* set, some component public_network is set, the cluster has a mix of listening on dedicated interfaces vs. all interfaces
- The global public_network is set, some component public_network is set, the cluster listens on the global public interface except for the components that listen on a different public_network interface

For the cluster_network, these two commands are redundant
` ceph config set global cluster_network`
`ceph config set osd cluster_network`
Any other component should be meaningless.

----
My thought is that we should be recommending something.  See #659 We are with the steps in this indirect way.  I suggest adding some version of the possibilities above to let the users know what is possible and then recommend the one solution such as setting the global public_network and global cluster_network.  (This does mean to not use the edit above where I replaced `global` with `osd`, but to also make corrections where `ceph config set osd` is used.  I can imagine the fun of somebody debugging where the global cluster_network and the osd cluster_network settings are different.)

smithfarm:
@swiftgist It's quite possible that "ceph config set global cluster_network" and "ceph config set osd cluster_network" are functionally equivalent, but exactly for the reason you stated - global trumps osd - I figure we had better recommend users to use "global", just like upstream does.

But if you insist on writing "osd" here, instead of "global", I won't object further.

swiftgist:
> @swiftgist It's quite possible that "ceph config set global cluster_network" and "ceph config set osd cluster_network" are functionally equivalent, but exactly for the reason you stated - global trumps osd - I figure we had better recommend users to use "global", just like upstream does.
> 
> But if you insist on writing "osd" here, instead of "global", I won't object further.

No, this is more the case of when I was investigating the other issue #659 and correcting that command that I hadn't thought about the consequences until seeing this section.  If we are/will be recommending that both global public_network and global cluster_network be set, then I think it could be stated more plainly.  Recommend the two commands and mention the ramifications of not setting either.  It's a toss up on mentioning the other possible uses for the component public_network, but since this is a deployment guide, this place seems as good as any.

xml/deploy_cephadm.xml
asettle: Going to remove it completely so it simply says "The next steps involve using the &cephadm; to deploy additional &mon;,  &mgr;, OSDs, the Monitoring Stack, and Gateways." that's all the customer needs to know

xml/deploy_cephadm.xml
asettle: We go with upstream for now, unless we need someone else to verify this?

asettle:
@swiftgist @smithfarm Not to derail what is a potentially important conversation but I have 6 days to ensure the documentation works. If there is an issue or a bug (or a worldly debate) it needs to happen elsewhere. My entire remit right now is to ensure that the documentation _works_ and that is it. I will push up the edits that @swiftgist has suggested, but provided the content is sane, I will be merging and closing the related bug reports.

lproven (approved):

xml/deployment_hwrecommend.xml
```diff
@@ -154,7 +154,7 @@
       <para>
        Set the cluster network using the following command:
       </para>
-<screen>&prompt.root;ceph config set osd cluster_network <replaceable>MY_NETWORK</replaceable></screen>
+<screen>&prompt.root;ceph config set global public_network <replaceable>MY_NETWORK</replaceable></screen>
```
asettle, line 157: Actually, this should be `global cluster_network`, no? @smithfarm

smithfarm: Yes

PR #673, opened 2020-10-08
Container registry systemd unit  file (#513)

### LINKED ISSUE #513: [doc] 6.3.2.10  Configure Container Registry
### Type of report
   - [x] bug
   - [ ] feature request

### Source Document
6.3.2.10  Configure Container Registry:

https://susedoc.github.io/doc-ses/master/single-html/ses-deployment/#deploy-cephadm-configure-registry

### Summary

`podman --restart=always` doesn't survive a reboot. 

We need to properly create a systemd service file for this to work.

The systemd unit file we document should be similar to the one generated by `podman generate systemd registry`

note that we should **NOT**  recommend users to run `podman generate systemd registry` by themselves.

tbazant:
closes #513

### DIFF
--- a/xml/deploy_cephadm.xml
+++ b/xml/deploy_cephadm.xml
@@ -726,8 +726,8 @@ o- time_server ................................................ [enabled]
     <procedure>
      <step>
       <para>
-       Create a local container registry instead of the default public one. Create it
-       outside of &ceph; cluster nodes. The local registry needs to be accessible
+       Create a local container registry outside of the &ceph; cluster nodes instead
+       of the default public one as the local registry needs to be accessible
        from cluster nodes. Use a command similar to the following:
       </para>
 <screen>
--- a/xml/deploy_cephadm.xml
+++ b/xml/deploy_cephadm.xml
@@ -742,7 +742,7 @@ o- time_server ................................................ [enabled]
      </step>
      <step>
       <para>
-       Enable and start the registry service, then verify its status:
+       Enable and start the registry service, and then verify the status:
       </para>
 <screen>
 &prompt.root;systemctl enable suse_registry.service
--- a/xml/deploy_cephadm.xml
+++ b/xml/deploy_cephadm.xml
@@ -726,8 +726,9 @@ o- time_server ................................................ [enabled]
     <procedure>
      <step>
       <para>
-       Create a local container registry outside &ceph; cluster nodes that is
-       accessible from cluster nodes, for example:
+       Create a local container registry instead of the default public one. Create it
+       outside of &ceph; cluster nodes. The local registry needs to be accessible
+       from cluster nodes. Use a command similar to the following:
       </para>
 <screen>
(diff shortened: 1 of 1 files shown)


xml/deploy_cephadm.xml
```diff
@@ -726,11 +726,28 @@ o- time_server ................................................ [enabled]
     <procedure>
      <step>
       <para>
-       Create a local container registry accessible but outside of the &ceph;
-       cluster, for example:
+       Create a local container registry outside &ceph; cluster nodes that is
+       accessible from cluster nodes, for example:
```
asettle, line 730: I'm not sure I understand this better?

xml/deploy_cephadm.xml
```diff
@@ -726,11 +726,28 @@ o- time_server ................................................ [enabled]
...
+     <step>
+      <para>
+       Create the corresponding &systemd; unit file for easier registry operation:
+      </para>
+<screen>&prompt.root;podman generate systemd registry >  \
+ /etc/systemd/system/suse_registry.service</screen>
+     </step>
+     <step>
+      <para>
+       Enable and start the registry service, then verify its status:
```
asettle, line 745:
```suggestion
       Enable and start the registry service, and then verify the status:

File: xml/deploy_cephadm.xml (hunk 1 above)
Comment (tbazant) [inline xml/deploy_cephadm.xml:730]: i know....1 sentence is not enough
```suggestion
       Create a local container registry instead of the default public one. Create it
       outside of &ceph; cluster nodes. The local registry needs to be accessible
       from cluster nodes. Use a command similar to the following:
```

tbazant (commented):
just 1 suggestion

xml/deploy_cephadm.xml
asettle: Yeah that's better @tbazant - apply that :)

jschmid1:
That does start/enable _a_ registry. Is that the registry SUSE is offering?

xml/deploy_cephadm.xml
```diff
@@ -726,11 +726,29 @@ o- time_server ................................................ [enabled]
     <procedure>
      <step>
       <para>
-       Create a local container registry accessible but outside of the &ceph;
-       cluster, for example:
+       Create a local container registry instead of the default public one. Create it
+       outside of &ceph; cluster nodes. The local registry needs to be accessible
+       from cluster nodes. Use a command similar to the following:
```
asettle, line 731:
```suggestion
       Create a local container registry outside of the &ceph; cluster nodes instead
       of the default public one as the local registry needs to be accessible
       from cluster nodes. Use a command similar to the following:

asettle: This is just a suggestion, no need if you prefer your own version.

asettle (approved):

asettle:
> That does start/enable _a_ registry. Is that the registry SUSE is offering?

@tbazant is this verified?

PR #679, opened 2020-10-09
RBD mirror prompts fixed, minor fixes to text done (#677)

### LINKED ISSUE #677: For RBD mirror section, use illustrative prompts
In https://susedoc.github.io/doc-ses/master/single-html/ses-admin/#ceph-rbd-mirror

Would like to see these in the examples:
```
cephadm@local >
```
and for remote

```
cephadm@remote >
```

Rather than:

```
cephadm@adm
```

tbazant:
Replaced neutral prompts with '@local', '@remote' suffixes for RBD mirroring content.
@trociny @denisok if you find another occurrence worth updating, make a list and let me know please

closes #677

### DIFF
--- a/xml/admin_rbd.xml
+++ b/xml/admin_rbd.xml
@@ -947,56 +947,75 @@ data blocks changed from 2097152 to 2560000</screen>
        peer bootstrap create</command> command, a pool name, along with an optional
        friendly site name to describe the <literal>local</literal> cluster:
      </para>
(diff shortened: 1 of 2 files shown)


asettle (approved):

lproven (approved):

trociny (approved):
LGTM

Though as I already said it somewhere, I would prefer if we used "site-A/site-B" instead of "local /remote", as the upstream doc does. Because with the two-way mirroring "local /remote" does not make much sense.

PR #681, opened 2020-10-09
Add para/remark elements to avoid future problems

tomschr:
This PR contains the following change:

It was an oversight to allow "empty" sections in GeekoDoc. To avoid validation issues in the future, this commit introduces
para/remark lines.

Also remove literal in screen as it is useless. ;)


asettle (approved):

tbazant (approved):
looks good

PR #690, opened 2020-10-13
Updating certificate documentation for clarity (BSC#1175881)

asettle:


asettle:
@swiftgist this is just a really simple way to indicate that you can't have both. I think it would be fine this way, what do you think?

swiftgist:
I think the description is fine telling the user to pick one or the other.  Changing the append text to "If you have a valid certificate signed by a CA" might confuse those that are doing a self-signed cert.  That is, the self-signed group might not think that step applies to them now.

What if Step 0 (or renumbered) says "If you have a valid certificated signed by a CA, skip to Step 4"?

tbazant (approved):
lgtm

lproven (dismissed):
Github will not let me comment or make suggestions. I don't know why not.

I would like to propose 3 small changes to the existing text from line 1513:

«
   to either have a CA-issued certificate, or create a self-signed one&mdash;
   not both. There are two ways to configure &ogw; with HTTPS enabled: a simple
   way that makes use of the default settings, and an advanced way that lets you fine-
   tune HTTPS related settings.
»

• Hyphenate CA-issued
• Replace em-dash after "enabled" with a simple colon
• Hyphenate fine-tune

swiftgist (approved):

PR #704, opened 2020-10-19
Adding instructions about dense nodes (BSC#1177776)

asettle:
Requires release note

### DIFF
--- a/xml/deploy_cephadm.xml
+++ b/xml/deploy_cephadm.xml
@@ -1198,7 +1198,7 @@ config: OK
      make sure to assign sufficient ports for &ceph; OSDs. The default (6800-7300)
      currently allows for no more than 62 OSDs per host. For a cluster with dense
      nodes, adjust the setting <literal>ms_bind_port_max</literal> to a
-     suitable value. Each OSD will consume 8 additional ports. For example,
+     suitable value. Each OSD will consume eight additional ports. For example,
      given a host that is set to run 96 OSDs, 768 ports will be needed.
      <literal>ms_bind_port_max</literal> should be set at least to 7568 by running
      the following command:
--- a/xml/deploy_cephadm.xml
+++ b/xml/deploy_cephadm.xml
@@ -1193,6 +1193,17 @@ config: OK
      deployment will use the intended cluster network from the start.
     </para>
    </note>
+   <para>
+     If your cluster is set to have dense nodes (greater than 60 OSDs per host),
+     make sure to assign sufficient ports for &ceph; OSDs. The default (6800-7300)
+     currently allows for no more than 62 OSDs per host. For a cluster with dense
+     nodes, adjust the setting <literal>ms_bind_port_max</literal> to a
+     suitable value. Each OSD will consume 8 additional ports. For example,
+     given a host that is set to run 96 OSDs, 768 ports will be needed.
+     <literal>ms_bind_port_max</literal> should be set at least to 7568 by running
+     the following command:
+   </para>
+<screen>&prompt.smaster;ceph config set osd.* ms_bind_port_max 7568</screen>
   </sect2>
  </sect1>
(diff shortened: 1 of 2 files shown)


asettle:
@sknorr this change will require an addition to the release nodes. This note should say (something to the effect of): _SUSE Enterprise Storage cannot deploy more than 62 OSDs per node. If you intend to update a cluster with very dense nodes (>62 OSDs per node) read <link to doc section> before proceeding._ 

I believe putting it under Known ISsues would be suitable. All okay with you @jan--f ?

jan--f:
> @sknorr this change will require an addition to the release nodes. This note should say (something to the effect of): _SUSE Enterprise Storage cannot deploy more than 62 OSDs per node. If you intend to update a cluster with very dense nodes (>62 OSDs per node) read `link to doc section` before proceeding._
> 
> I believe putting it under Known ISsues would be suitable. All okay with you @jan--f ?

I'd like to propose a slight change to the release notes entry:
```
SUSE Enterprise Storage cannot deploy more than 62 OSDs per node in its default configuration. If you intend to update a cluster with very dense nodes (>62 OSDs per node) read `link to doc section` before proceeding.
```
_Known Issues_ section sounds alright.

xml/deploy_cephadm.xml
```diff
@@ -1193,6 +1193,17 @@ config: OK
      deployment will use the intended cluster network from the start.
     </para>
    </note>
+   <para>
+     If your cluster is set to have dense nodes (greater than 60 OSDs per host),
```
jan--f, line 1197:
```suggestion
     If your cluster is set to have dense nodes (greater than 62 OSDs per host),
```
Just for consistency its probably better to use the same number? Apologies, should have thought of that in the initial proposal.

jan--f (approved):
One minor nit, but otherwise LGTM!

jan--f (approved):
:+1:

xml/deploy_cephadm.xml
```diff
@@ -1193,6 +1193,21 @@ config: OK
      deployment will use the intended cluster network from the start.
     </para>
    </note>
+   <para>
+     If your cluster is set to have dense nodes (greater than 62 OSDs per host),
+     make sure to assign sufficient ports for &ceph; OSDs. The default (6800-7300)
+     currently allows for no more than 62 OSDs per host. For a cluster with dense
+     nodes, adjust the setting <literal>ms_bind_port_max</literal> to a
+     suitable value. Each OSD will consume 8 additional ports. For example,
```
lproven, line 1201:
```suggestion
     suitable value. Each OSD will consume eight additional ports. For example,
```

xml/deploy_cephadm.xml
```diff
@@ -1193,6 +1193,21 @@ config: OK
      deployment will use the intended cluster network from the start.
     </para>
    </note>
+   <para>
+     If your cluster is set to have dense nodes (greater than 62 OSDs per host),
+     make sure to assign sufficient ports for &ceph; OSDs. The default (6800-7300)
```
lproven, line 1198: I am not sure it is, but if this is a _range_ of ports, then I think "the default range (6800-7300)" would be preferable.

lproven (changes requested):

xml/deploy_cephadm.xml
asettle: Good question, I presume it's a range based on how we've phrased this here: https://susedoc.github.io/doc-ses/master/html/ses-troubleshooting/storage-tips.html#storage-bp-net-firewall

@jan--f ?

jan--f: Yep makes sense!

asettle, line 1198:
```suggestion
     make sure to assign sufficient ports for &ceph; OSDs. The default range (6800-7300)

lproven (approved):

Issue #698, opened 2020-10-16
Document config-key store


tbazant:
enhance this chapter https://susedoc.github.io/doc-ses/master/html/ses-admin/cha-ceph-configuration.html and include:
* general information about using `ceph config-key` to store and retrieve saved values for ceph application
* description of config-key store keys for specific application, such as ceph-iscsi in #669

tbazant:
Can you contribute with key for RGW, NFS, RBD, RBD-MIRROR etc if applicable?
@theanalyst @trociny @dthpulse @ddiss @jschmid1 @matthewoliver

trociny:
rbd and  rbd-mirror may use config-key store for peer configuration, but one should use (higher level) `rbd` cli to update this configuration, which is documented in rbd/rbd-mirror documentation. So I think nothing rbd/rbd-mirror specific should be mentioned in this section: users are expected to use `rbd` cli and not low level `ceph config-key` for this.

PR #724, opened 2020-10-26
Added config-key store documentation (#698)

### LINKED ISSUE #669: ceph-iscsi config options in mon-store
Only after https://github.com/ceph/ceph/pull/37588 merges:

```Currently the ceph-iscsi config options get placed in the
iscsi-gateway.cfg but this is volume mounted into the container and
lives under /var/lib/ceph/<fsid>/<iscsi service>/iscsi-gateway.cfg. This is
fine for day 1 situations but horrible on day 2.

To make this easier a PR was pushed to ceph-iscsi[0] which taught it how
to look in the mon config-key store for values to config options. This
took the same approach as RGW. A config value starting with a
config:// will be treated as key in the mon store. E.g:

config://my/awesome/key

Will go look for the value of my/awesome/key.```

This patch adds a new option to the iscsi spec called mon_key_prefix
which defaults to iscsi. Which allows someone to deploy a ceph-iscsi
service with either globally shared options or can seperate as required.

### PR DESCRIPTION
**Pull request by tbazant**
closes #669 #698 

rendered version is attached 
[cha-ceph-config-key-store_color_en.pdf](https://github.com/SUSE/doc-ses/files/5439159/cha-ceph-config-key-store_color_en.pdf)

### CODE REVIEW

File: xml/runtime_config.xml [hunk 1]
```diff
@@ -294,11 +294,101 @@
    </sect3>
   </sect2>
  </sect1>
- <sect1 xml:id="ceph-config-key-store">
+ <sect1 xml:id="cha-ceph-config-key-store">
   <title>Config-key Store</title>
   <para>
-   Config-key store is a general purpose key/value service offered by the monitors.
+   Config-key is a general purpose service offered by the &mon;s. It simplifies
```
lproven, line 300: Is `config-key` a command, or a program? I think when it is first introduced, it should be emphasized in some way, e.g. [systemitem class=service] or something.

lproven, line 300:
```suggestion
   Config-key is a general-purpose service offered by the &mon;s. It simplifies
```

lproven (changes requested):

xml/runtime_config.xml
tbazant: @lproven made it an entity, have a look once more please

lproven (approved):

PR #783, opened 2020-11-19
Improving description and procedure for EC pools

### LINKED ISSUE #775: [doc] 10.1.2  Creating a Block Device Image in an Erasure Coded Pool
### Type of report
   - [ ] bug
   - [ ] feature request

### Source Document
10.1.2  Creating a Block Device Image in an Erasure Coded Pool:

https://documentation.suse.com/ses/7/single-html/ses-admin/#ceph-rbd-cmds-create-ec

### Summary
I would like we improve description about what the metadata pool is. In the current version it might be confusing.

I would add, that the metadata pool is the pool which you should use when referring the image (i.e.`rbd info $pool/$image`) and `rbd list` will list the image in the metadata pool. And the sentence: "You need to specify the replicated pool for storing the image's metadata with the --pool= option of the rbd create command." not entirely correct, because you can also specify it as "pool/" prefix to the image name.

And then I would change the command:
```
rbd create IMAGE_NAME --size=1G --data-pool POOL_NAME --pool=OTHER_POOL
```
to:
```
rbd create IMAGE_NAME --size=1G --data-pool EC_POOL --pool=POOL
```
or:
```
rbd create POOL/IMAGE_NAME --size=1G --data-pool EC_POOL 
```

asettle:
Closes #775

### DIFF
--- a/xml/admin_rbd.xml
+++ b/xml/admin_rbd.xml
@@ -100,8 +100,8 @@
    <para>
     Create an EC pool:
    </para>
-<screen>&prompt.cephuser;ceph osd pool create <replaceable>POOL_NAME</replaceable> 12 12 erasure
-&prompt.cephuser;ceph osd pool set <replaceable>POOL_NAME</replaceable> allow_ec_overwrites true</screen>
+<screen>&prompt.cephuser;ceph osd pool create <replaceable>EC_POOL</replaceable> 12 12 erasure
+&prompt.cephuser;ceph osd pool set <replaceable>EC_POOL</replaceable> allow_ec_overwrites true</screen>
    <para>
      Specify the replicated pool for storing metadata:
    </para>
@@ -114,10 +114,6 @@
 <screen>
 &prompt.cephuser;rbd create <replaceable>POOL/IMAGE_NAME</replaceable> --size=1G --data-pool EC_POOL
 </screen>
-   <para>
-     Use the following command to list the image of the metadata pool:
-   </para>
-<screen>&prompt.cephuser;rbd list</screen>
   </sect2>
 
   <sect2 xml:id="ceph-rbd-cmds-list">
--- a/xml/admin_rbd.xml
+++ b/xml/admin_rbd.xml
@@ -84,27 +84,40 @@
   <sect2 xml:id="ceph-rbd-cmds-create-ec">
    <title>Creating a Block Device Image in an Erasure Coded Pool</title>
    <para>
-    As of &productname; 5, it is possible to store data of a block device image
+    It is possible to store data of a block device image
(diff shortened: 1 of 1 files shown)


xml/admin_rbd.xml
```diff
@@ -84,27 +84,40 @@
...
    <para>
-    Use the following steps to create an RBD image in a newly created EC pool:
+    Create an EC pool:
    </para>
-<screen>&prompt.cephuser;<command>ceph</command> osd pool create <replaceable>POOL_NAME</replaceable> 12 12 erasure
-&prompt.cephuser;<command>ceph</command> osd pool set <replaceable>POOL_NAME</replaceable> allow_ec_overwrites true
-
-#Metadata will reside in pool "<replaceable>OTHER_POOL</replaceable>", and data in pool "<replaceable>POOL_NAME</replaceable>"
-&prompt.cephuser;<command>rbd</command> create <replaceable>IMAGE_NAME</replaceable> --size=1G --data-pool <replaceable>POOL_NAME</replaceable> --pool=<replaceable>OTHER_POOL</replaceable></screen>
+<screen>&prompt.cephuser;ceph osd pool create <replaceable>POOL_NAME</replaceable> 12 12 erasure
```
trociny, line 103: Just thinking. Probably we want to use `EC_POOL` here too (and in the command below), so it is the same we use in the `rbd create` examples.

Or some other name you like more, just make sure it is the same in all examples and the name not so generic as it was previously (`POOL_NAME`).

xml/admin_rbd.xml
```diff
@@ -84,27 +84,40 @@
...
+&prompt.cephuser;rbd create <replaceable>IMAGE_NAME</replaceable> --size=1G --data-pool <replaceable>EC_POOL</replaceable> --pool=<replaceable>POOL</replaceable>
+</screen>
+   <para>
+     Or:
+   </para>
+<screen>
+&prompt.cephuser;rbd create <replaceable>POOL/IMAGE_NAME</replaceable> --size=1G --data-pool EC_POOL
+</screen>
+   <para>
+     Use the following command to list the image of the metadata pool:
```
trociny, line 118: Hm. I am not sure I like it. What I meant that when you run `rbd list $pool` or `rbd list --pool $pool` it will show images created with metadata pool `$pool`.

Actually, after your changes above, we probably don't need this paragraph at all. It looks clear enough.

trociny (commented):
otherwise LGTM

xml/admin_rbd.xml
asettle: Will remove it :)

xml/admin_rbd.xml
asettle: Done!

trociny (approved):

PR #790, opened 2020-11-20
windows: Fix capitalization of Dokany

ml8mr:
The upstream Dokany project is capitalized, so we should
use that in our docs as well.

Signed-off-by: Mike Latimer <mlatimer@suse.com>


tbazant (approved):
lgtm

PR #799, opened 2020-11-25
Adding insecure=true to ceph-salt command for local container registry

ghost:
This adds one option, insecure=true, to the ceph-salt command for registration of a local container registry.
With this modification, insecure=true is added to /etc/containers/registries.conf, allowing you to connect to the container registry without TLS.
TLS are disabled on the container registry deployed in the way described in the SES7 deployment guide.
So you fail to retrieve the container images without this option.

Enabling TLS on the container registry may be another solution.
But the guide uses --dest-tls-verify=false when mirroring the container images with skopeo, so I think that it is not intended to use TLS connections.

### DIFF
--- a/xml/deploy_cephadm.xml
+++ b/xml/deploy_cephadm.xml
@@ -841,7 +841,8 @@ o- time_server ................................................ [enabled]
 <screen>
 &prompt.smaster;ceph-salt config /containers/registries_conf/registries \
  add prefix=registry.suse.com \
- location=<replaceable>LOCAL_REGISTRY_HOST_IP</replaceable>:5000/registry.suse.com
+ location=<replaceable>LOCAL_REGISTRY_HOST_IP</replaceable>:5000/registry.suse.com \
+ insecure=true
 </screen>
      </step>
     </procedure>


asettle:
@knkonishi I believe we've discussed this at length internally, and we have removed this option for varied reasons. I will be bringing some people in to review.

@sebastian-philipp @smithfarm @jan--f

smithfarm:
Not me, but maybe @Martin-Weiss ?

Martin-Weiss:
IMO "insecure=true" is a bug in design, deployment and configuration.. We MUST deliver products in a secure way and this means we have to enable SSL on the on-premise registry with proper certificates and we have to trust the CA used for the certificate creation. Please do NOT configure and document to use insecure=true per default !

P.S. an on-premise registry is also expected to allow authenticated and authorized access, only..

knkonishi:
Thank you for your response.
I see that the insecure option is not recommened.
But the document says the following:

https://documentation.suse.com/ses/7/html/ses-all/deploy-cephadm.html#deploy-cephadm-configure-registry
> Creating a local registry is useful in the following scenarios:
> ...
> If configuration or network issues prevent your cluster from accessing remote registries across a secure link, so you need a local, unencrypted registry instead.

This seems to me that an insecure local registry is acceptable.
If not, could you add tips about SSL certificates to the document?

Martin-Weiss:
> Thank you for your response.
> I see that the insecure option is not recommened.
> But the document says the following:
> 
> https://documentation.suse.com/ses/7/html/ses-all/deploy-cephadm.html#deploy-cephadm-configure-registry
> 
> > Creating a local registry is useful in the following scenarios:
> > ...
> > If configuration or network issues prevent your cluster from accessing remote registries across a secure link, so you need a local, unencrypted registry instead.
> 

I believe this is just wrong.. „unencrypted“ might also be something different than „https“.. however - if „unencrypted means clear text“ - then this is wrong.

> This seems to me that an insecure local registry is acceptable.

Technically this might be possible somehow - but for production deployments it would be wrong as it would not fulfill customers security and audit requirements.

> If not, could you add tips about SSL certificates to the document?

There are several documents available that explain how to setup a on-premise registry with the required SSL and with authentication / authorization.. 

I.e. this one
https://documentation.suse.com/suse-caasp/4.5/single-html/caasp-airgap/#airgap-container_registry

Or here - but you need to follow the registry configuration documentation on how to add SSL
https://documentation.suse.com/sles/15-SP2/html/SLES-all/cha-registry-installation.html

Also in the Portus documentation http://port.us.org/docs/deploy.html has some explanations 

Basically the process is 
- create ssl certificate and key
- add the right start options to the container start for the docker-registry and to portus

I will add an other comment in a few minutes with some container start examples..

Martin-Weiss:
Just pushed some example scripts for registry with portus and using SSL to https://github.com/Martin-Weiss/registry - HTH.

asettle:
Hi @knkonishi and @Martin-Weiss - I am going to go ahead and close this issue. We need to internally discuss the best way to configure a local registry. This pull request will not be deleted, simply closed. The comments will remain for future reference.

@knkonishi - we are working on this. Thank you for bringing this up :) I will ping you when we have a change in the documentation for you to review. Is this suitable?

Thank you,

Alex

knkonishi:
@Martin-Weiss and @asettle - Thank you for prividing a lot of helpful information.

@asettle 
> I will ping you when we have a change in the documentation for you to review. Is this suitable?

Yes, no problem.
I would appreciate it if you could update the relevant parts of the document as soon as possible.

sebastian-philipp:
@knkonishi https://jira.suse.com/browse/SES-2304

PR #803, opened 2020-11-25
Adding final step to SSL config for RGW

### LINKED ISSUE #744: Verify SSL needs to be disabled when configuring the RGW with a self signed certificate
**Link To Section**
https://documentation.suse.com/ses/6/single-html/ses-admin/#ogw-ssl-simple

**Summary**
If the RGW is configured to use SSL with a self signed certificate as per the above documentation, then it is also required to disable SSL verififcation else dashboard access will fail with a non descriptive error message simply stating:

`Sorry, we could not find what you were looking for`

**Steps To Reproduce**
Follow the documentation above to enable SSL access for the RGW.

**Expected Results**
Dashboard access should not fail after taking the above steps.

**Notes**
See also: https://bugzilla.suse.com/show_bug.cgi?id=1178355#c7
To resolve this, the following step needs to be added to the documentation:

Finally run:

`ceph dashboard set-rgw-api-ssl-verify False`

asettle:
Closes #744 

Requires forward-port to SES 7

### DIFF
--- a/xml/admin_ceph_gateway.xml
+++ b/xml/admin_ceph_gateway.xml
@@ -726,6 +726,12 @@ rgw_init: default-ssl
 &prompt.smaster;salt-run state.orch ceph.stage.4
 </screen>
     </step>
+    <step>
+      <para>
+        Finally, run:
+      </para>
+<screen>&prompt.smaster;ceph dashboard set-rgw-api-ssl-verify False</screen>
+     </step>
    </procedure>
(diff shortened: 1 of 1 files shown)


lproven (approved):

lproven (approved):

tbazant (approved):
sounds good

theanalyst (approved):

PR #804, opened 2020-11-25
Removing indepth section and replacing with support info

### LINKED ISSUE #771: [doc] 4.3.1  An OSD Will Not Start
### Type of report
   - [X] bug
   - [X] feature request

### Source Document
4.3.1  An OSD Will Not Start:

https://documentation.suse.com/ses/7/single-html/ses-troubleshooting/#troubleshooting-osd-start

### Summary
This documents how to adjust /etc/sysctl.conf, however we do deploy `/etc/sysctl.d/90-ceph-osd.conf` that we should modify.

However, we also have https://github.com/SUSE/ceph/blob/ses7/etc/sysctl/CMakeLists.txt which suggests to me we already include exactly this setting.

Maybe the entire section can just be dropped in the docs?

Further, I'm a bit confused what "_Kernel Version: Identify the kernel version and distribution you are using. Ceph uses some third party tools by default, which may be buggy or may conflict with certain distributions or kernel versions._" is asking one of our customers to do.

As for segmentation faults, these should result in a coredump and I think we should instead provide guidance on how to check if they've been generated and how to submit them to SUSE? Perhaps via supportconfig?

@smithfarm @jan--f, thoughts? @tserong?

asettle:
Closes #771

### DIFF
--- a/xml/bp_troubleshooting_osds.xml
+++ b/xml/bp_troubleshooting_osds.xml
@@ -159,41 +159,10 @@
   <sect2 xml:id="troubleshooting-osd-start">
     <title>An OSD Will Not Start</title>
     <para>
-      If you start your cluster and an OSD will not start, check the following:
-    </para>
-    <itemizedlist>
-      <listitem>
-        <para>
-          <emphasis>Check Max Threadcount</emphasis>: If you have a node with a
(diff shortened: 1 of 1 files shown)


tbazant (approved):
looks good

smithfarm (approved):

PR #807, opened 2020-11-26
Adding crush removal step

### LINKED ISSUE #745: [doc] 3.3  Removing Nodes
### Type of report
   - [X] bug
   - [ ] feature request

### Source Document
3.3  Removing Nodes:

https://susedoc.github.io/doc-ses/master/html/ses-admin/storage-salt-cluster.html#salt-node-removing

### Summary

From:

https://tracker.ceph.com/issues/48260
https://bugzilla.suse.com/show_bug.cgi?id=1176265

After removing all OSDs on a single host, the user has to manually remove the "host" from the crush map.

asettle:
Closes #745

### DIFF
--- a/xml/admin_saltcluster.xml
+++ b/xml/admin_saltcluster.xml
@@ -201,6 +201,17 @@ placement:
 <screen>&prompt.cephuser;ceph-salt config /ceph_cluster/roles/bootstrap reset</screen>
    </step>
    <step>
+    <para>
+      After removing all OSDs on a single host, remove the host from the crush map:
+    </para>
+<screen>&prompt.cephuser;ceph osd crush remove <replaceable>bucket-name</replaceable></screen>
+    <note>
+      <para>
+        The bucket name should be the same as the host name.
+      </para>
+     </note>
+   </step>
+   <step>
      <para>
        You can now remove the minion from the cluster:
      </para>
--- a/xml/admin_saltcluster.xml
+++ b/xml/admin_saltcluster.xml
@@ -202,7 +202,7 @@ placement:
    </step>
    <step>
     <para>
-      After removing all OSDs on a single host, remove the host from the crush map:
+      After removing all OSDs on a single host, remove the host from the CRUSH map:
     </para>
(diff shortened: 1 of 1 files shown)


xml/admin_saltcluster.xml
```diff
@@ -200,6 +200,17 @@ placement:
...
 <screen>&prompt.cephuser;ceph-salt config /ceph_cluster/roles/bootstrap reset</screen>
    </step>
+   <step>
+    <para>
+      After removing all OSDs on a single host, remove the host from the crush map:
+    </para>
+<screen>&prompt.cephuser;ceph osd crush remove <replaceable>bucket-name</replaceable></screen>
+    <note>
+      <para>
+        The bucket name should be the same as the host name.
```
asettle, line 210: @jecluis if you think we need to be more verbose here, let me know. It's a little vague, but I do feel by this step they should be aware of how to remove this. Might be too much assumed knowledge, however.

xml/admin_saltcluster.xml
```diff
@@ -200,6 +200,17 @@ placement:
     </para>
 <screen>&prompt.cephuser;ceph-salt config /ceph_cluster/roles/bootstrap reset</screen>
    </step>
+   <step>
+    <para>
+      After removing all OSDs on a single host, remove the host from the crush map:
```
lproven, line 205:
```suggestion
      After removing all OSDs on a single host, remove the host from the CRUSH map:
```

lproven: CRUSH is an acronym, isn't it?

lproven (changes requested):

lproven (approved):

xml/admin_saltcluster.xml
jecluis: I think this is clear enough. Unless ones does not read the surrounding text and just goes about running commands blindly, but I think we can hardly blame the docs for that.

jecluis (approved):

PR #825, opened 2020-12-08
Removing registry info (BSC#1177210)

### LINKED ISSUE #744: Verify SSL needs to be disabled when configuring the RGW with a self signed certificate

asettle:
This PR addresses the issues in BSC#1177210 by changing the following:

- Removes the podmon and skopeo local registry creation
- Removes the `<tip>` admonition and makes the private registry SBP article the main source of information: https://documentation.suse.com/sbp/all/single-html/SBP-Private-Registry/

Here is the build: 

Screenshot from 2020-12-08 15-34-32


asettle:
@sebastian-philipp  is going to create a a Confluence page with the old information on podman and skopeo including the details on how to setup with an insecure registry. This will remain INTERNAL ONLY as it will provide a suitable alternative to those who simply need to setup a registry for POCs.

ajaeger:
LGTM, thanks

PR #830, opened 2020-12-09
Updating the SES glossary

asettle:
XML formatting file: ses_glossary.xml ... successful

### DIFF
--- a/xml/ses_glossary.xml
+++ b/xml/ses_glossary.xml
@@ -29,7 +29,7 @@
    <glossdef>
     <para>
      A single binary which handles alerts sent by the &prometheus; server and
-     notifies end user.
+     notifies the end user.
     </para>
    </glossdef>
   </glossentry>
@@ -113,7 +113,7 @@
      algorithm that determines how to store and retrieve data by computing data
      storage locations. CRUSH requires a map of the cluster to pseudo-randomly
      store and retrieve data in OSDs with a uniform distribution of data across
-     the cluster.The threa
+     the cluster.
     </para>
    </glossdef>
   </glossentry>
--- a/xml/ses_glossary.xml
+++ b/xml/ses_glossary.xml
@@ -101,7 +101,7 @@
   <glossentry><glossterm>&ceph; Storage Cluster</glossterm>
    <glossdef>
     <para>
-     The core set of storage software which stores the user’s data. Such a set
+     The core set of storage software which stores the user's data. Such a set
      consists of &ceph; monitors and OSDs.
     </para>
    </glossdef>
@@ -166,7 +166,7 @@
   <glossentry><glossterm>&mgr;</glossterm>
    <glossdef>
     <para>
-     &mgr; or MGR is the &ceph; manager software, which collects all the state
+     &mgr; or MGR is the &ceph; Manager software, which collects all the state
      from the whole cluster in one place.
     </para>
    </glossdef>
@@ -207,7 +207,7 @@
   <glossentry><glossterm>&ceph; Object Storage</glossterm>
    <glossdef>
     <para>
-     The object storage “product”, service or capabilities, which consists of a
+     The object storage "product", service or capabilities, which consists of a
      &ceph; Storage Cluster and a &ceph; &ogw;.
     </para>
    </glossdef>
@@ -252,7 +252,7 @@
   <glossentry><glossterm>Reliable Autonomic Distributed Object Store (&rados;)</glossterm>
    <glossdef>
     <para>
-     The core set of storage software which stores the user’s data (MON+OSD).
+     The core set of storage software which stores the user's data (MON+OSD).
     </para>
    </glossdef>
   </glossentry>
--- a/xml/ses_glossary.xml
+++ b/xml/ses_glossary.xml
@@ -47,12 +47,6 @@
      A point that aggregates other nodes into a hierarchy of physical
      locations.
     </para>
-    <important>
-     <para>
-      Not to be confused with S3 <emphasis>buckets</emphasis> or
-      <emphasis>containers</emphasis>.
-     </para>
-    </important>
    </glossdef>
   </glossentry>
   <glossentry><glossterm>&cephadm;</glossterm>
@@ -83,7 +77,7 @@
    <glossdef>
     <para>
      The collection of &ceph; components which can access a &ceph; Storage
-     Cluster. These include the &ogw;, the Ceph Block Device, the &cephfs;, and
+     Cluster. These include the &ogw;, the &ceph; Block Device, the &cephfs;, and
      their corresponding libraries, kernel modules, and FUSEs.
     </para>
    </glossdef>
--- a/xml/ses_glossary.xml
+++ b/xml/ses_glossary.xml
@@ -125,7 +125,7 @@
   <glossentry><glossterm>&dashboard;</glossterm>
    <glossdef>
     <para>
-     A built-in web-based &ceph; management and monitoring application to
+     A built-in Web-based &ceph; management and monitoring application to
      administer various aspects and objects of the cluster. The dashboard is
      implemented as a &mgr; module.
     </para>
@@ -164,7 +164,7 @@
   <glossentry><glossterm>&mgr;</glossterm>
    <glossdef>
     <para>
-     &mgr; or MGR is the &ceph; Manager software, which collects all the state
+     &mgr; or MGR is the &ceph; manager software, which collects all the state
      from the whole cluster in one place.
     </para>
    </glossdef>
@@ -172,7 +172,7 @@
   <glossentry><glossterm>Node</glossterm>
    <glossdef>
     <para>
-     Any single machine or server in a Ceph cluster.
+     Any single machine or server in a &ceph; cluster.
     </para>
    </glossdef>
   </glossentry>
--- a/xml/ses_glossary.xml
+++ b/xml/ses_glossary.xml
@@ -48,11 +48,9 @@
      locations.
     </para>
     <important>
-     <title>Do not mix with S3 buckets</title>
      <para>
-      S3 <emphasis>buckets</emphasis> or <emphasis>containers</emphasis>
-      represent different terms meaning <emphasis>folders</emphasis> for
-      storing objects.
+      Not to be confused with S3 <emphasis>buckets</emphasis> or
+      <emphasis>containers</emphasis>.
      </para>
     </important>
    </glossdef>
--- a/xml/ses_glossary.xml
+++ b/xml/ses_glossary.xml
@@ -78,7 +78,7 @@
     <para>
      The collection of &ceph; components which can access a &ceph; Storage
      Cluster. These include the &ogw;, the &ceph; Block Device, the &cephfs;, and
-     their corresponding libraries, kernel modules, and FUSEs.
+     their corresponding libraries, kernel modules, and FUSE clients.
     </para>
    </glossdef>
   </glossentry>
@@ -112,7 +112,7 @@
   <glossentry><glossterm>CRUSH rule</glossterm>
    <glossdef>
     <para>
-     The CRUSH data placement rule that applies to a particular pool(s).
+     The CRUSH data placement rule that applies to a particular pool or pools.
     </para>
    </glossdef>
   </glossentry>
--- a/xml/ses_glossary.xml
+++ b/xml/ses_glossary.xml
@@ -25,6 +25,22 @@
     </para>
    </glossdef>
   </glossentry>
+  <glossentry><glossterm>&alertmanager;</glossterm>
+   <glossdef>
+    <para>
+     A single binary which handles alerts sent by the &prometheus; server and
+     notifies end user.
+    </para>
+   </glossdef>
+  </glossentry>
+  <glossentry><glossterm>archive sync module</glossterm>
+   <glossdef>
+    <para>
+     Module that enables creating an &ogw; zone for keeping the history of S3
+     object versions.
+    </para>
+   </glossdef>
+  </glossentry>
   <glossentry><glossterm>Bucket</glossterm>
    <glossdef>
     <para>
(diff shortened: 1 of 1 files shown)


xml/ses_glossary.xml
```diff
@@ -25,14 +25,30 @@
     </para>
    </glossdef>
   </glossentry>
+  <glossentry><glossterm>&alertmanager;</glossterm>
+   <glossdef>
+    <para>
+     A single binary which handles alerts sent by the &prometheus; server and
+     notifies end user.
```
lproven, line 32: Either:

... notifies the end-user.

Or

... notifies end-users.

xml/ses_glossary.xml
```diff
@@ -41,31 +57,139 @@
...
+  </glossentry>
   <glossentry><glossterm>CRUSH, &crushmap;</glossterm>
    <glossdef>
     <para>
      <emphasis>Controlled Replication Under Scalable Hashing</emphasis>: An
      algorithm that determines how to store and retrieve data by computing data
      storage locations. CRUSH requires a map of the cluster to pseudo-randomly
      store and retrieve data in OSDs with a uniform distribution of data across
-     the cluster.
+     the cluster.The threa
```
lproven, line 116: There looks to be a fragment of another sentence on the end of this line...?

lproven (changes requested):

xml/ses_glossary.xml
```diff
@@ -41,6 +57,55 @@
...
+    <para>
+     Provides tooling for deploying &ceph; clusters managed by &cephadm; using
+     Salt.
+    </para>
+   </glossdef>
+  </glossentry>
+  <glossentry><glossterm>&ceph; Storage Cluster</glossterm>
+   <glossdef>
+    <para>
+     The core set of storage software which stores the user’s data. Such a set
```
lproven, line 104:
```suggestion
     The core set of storage software which stores the user's data. Such a set
```

xml/ses_glossary.xml
```diff
@@ -52,20 +117,79 @@
     </para>
    </glossdef>
   </glossentry>
-  <glossentry><glossterm>Monitor node, MON</glossterm>
+  <glossentry><glossterm>CRUSH rule</glossterm>
+   <glossdef>
+    <para>
+     The CRUSH data placement rule that applies to a particular pool(s).
```
lproven, line 123: TBH I don't like the a...pools interpretation here but I feel maybe I'm being overly picky.

xml/ses_glossary.xml
```diff
@@ -52,20 +117,79 @@
...
+     &mon; or MON is the &ceph; monitor software.
+    </para>
+   </glossdef>
+  </glossentry>
+  <glossentry><glossterm>&mgr;</glossterm>
    <glossdef>
     <para>
-     A cluster node that maintains maps of cluster state, including the monitor
-     map, or the OSD map.
+     &mgr; or MGR is the &ceph; manager software, which collects all the state
```
lproven, line 169:
```suggestion
     &mgr; or MGR is the &ceph; Manager software, which collects all the state
```

xml/ses_glossary.xml
```diff
@@ -80,10 +204,19 @@
     </para>
    </glossdef>
   </glossentry>
-  <glossentry><glossterm>Node</glossterm>
+  <glossentry><glossterm>&ceph; Object Storage</glossterm>
    <glossdef>
     <para>
-     Any single machine or server in a Ceph cluster.
+     The object storage “product”, service or capabilities, which consists of a
```
lproven, line 210:
```suggestion
     The object storage "product", service or capabilities, which consists of a
```

xml/ses_glossary.xml
```diff
@@ -52,20 +117,79 @@
...
    </glossdef>
   </glossentry>
   <glossentry><glossterm>OSD</glossterm>
    <glossdef>
     <para>
-     Depending on context, <emphasis>Object Storage Device</emphasis> or
-     <emphasis>Object Storage Daemon</emphasis>. The
-     <command>ceph-osd</command> daemon is the component of &ceph; that is
+     <emphasis>Object Storage Device</emphasis>: A physical or logical storage
+     unit.
```
lproven, line 185: Are we deprecating the "daemon" meaning?

xml/ses_glossary.xml
lproven: Be careful of automatically-inserted smart/curly quotes!

xml/ses_glossary.xml
```diff
@@ -95,82 +228,81 @@
...
-  <glossentry><glossterm>&ceph; Storage Cluster</glossterm>
+  <glossentry><glossterm>Reliable Autonomic Distributed Object Store (&rados;)</glossterm>
    <glossdef>
     <para>
-     The core set of storage software which stores the user’s data. Such a
-     set consists of &ceph; monitors and OSDs.
-    </para>
-    <para>
-     AKA <quote>&ceph; Object Store</quote>.
+     The core set of storage software which stores the user’s data (MON+OSD).
```
lproven, line 255:
```suggestion
     The core set of storage software which stores the user's data (MON+OSD).
```

xml/ses_glossary.xml
```diff
@@ -25,14 +25,30 @@
...
+  </glossentry>
   <glossentry><glossterm>Bucket</glossterm>
    <glossdef>
     <para>
      A point that aggregates other nodes into a hierarchy of physical
      locations.
     </para>
     <important>
-     <title>Do Not Mix with S3 Buckets</title>
+     <title>Do not mix with S3 buckets</title>
```
lproven, line 51: By "mix" do we mean "confuse" here?

If the purpose of this para is to disambiguate S3 buckets from some other thing named "buckets" then I think maybe we should specify what that other thing is.

If we are disambiguating containers from (e.g.) Docker or k8s containers, then we should separate the entries for "bucket" and "container".

lproven (changes requested):
As a general comment, when substantially changing a variable-list or glossary, it makes for a very confusing diff if you edit term definitions into definitions of totally different terms. Could I ask that in future insert new entries for new terms, and amend existing ones only if the term is remaining substantially the same and only the definition is changing?

asettle:
> As a general comment, when substantially changing a variable-list or glossary, it makes for a very confusing diff if you edit term definitions into definitions of totally different terms. Could I ask that in future insert new entries for new terms, and amend existing ones only if the term is remaining substantially the same and only the definition is changing?

Ah yes, totally understandable! Will do my best to edit it differently next time :) or, would a rendered option suffice?

xml/ses_glossary.xml
asettle: I'm actually not sure, this was what you had previously?

xml/ses_glossary.xml
asettle: What would you recommend?

xml/ses_glossary.xml
asettle: It's now below :)

xml/ses_glossary.xml
asettle: True!

xml/ses_glossary.xml
asettle: Sorry, I am sure: It's "do not confuse" - I don't know why I didn't read that the right way the first time. But I think that tells you that it's confusing as it is...

xml/ses_glossary.xml
```diff
@@ -25,22 +25,85 @@
...
      locations.
     </para>
     <important>
-     <title>Do Not Mix with S3 Buckets</title>
      <para>
-      S3 <emphasis>buckets</emphasis> or <emphasis>containers</emphasis>
-      represent different terms meaning <emphasis>folders</emphasis> for
-      storing objects.
+      Not to be confused with S3 <emphasis>buckets</emphasis> or
+      <emphasis>containers</emphasis>.
```
lproven, line 53: OK, reword appreciated, but TBH I still do not feel it clearly disambiguates the term(s).

So, for example, based on 15min on Google, reading various bits of docs including ours, Amazon's, OpenStack's and others, I very tentatively suggest something like:

«
Be careful not to confuse the SES meanings of "buckets" and "containers" with uses of these terms by other products. 

OGW buckets are lists of OSDs, used to collect devices into physical locations. This is entirely distinct from the term as used in the context of Amazon S3.

S3 buckets are closer to what are termed containers in the context of OpenStack Swift. Here, containers are the entities that store objects, and must not be confused with operating system containers, for example those managed by Docker or CRIO.
»

This is based on my very fragmentary understanding.

What I am trying to get at is that to successfully make the distinction clear between "word X in context 1" versus "word X in context 2", we need to spell out _both_ contexts and explain the difference.

Worse still is what we seem to have here: where we _also_ have word Y in context 3 and word Y in context 4, and Word X.1 is the same **word** as Word X.2 but actually what it _means_ is the same as word Y.2.

If I have this right: 

An OpenStack Container is similar to an S3 Bucket. This is totally different from an OGW Bucket or a Docket Container.

Please don't take my word for it, though. All I am saying is that if we have buckets (2 different meanings) and containers (2 different meanings), we need to explicitly say:

"Bucket" in our context means "container" in their context. That does not mean the same as "container" in THAT context.

I submit that these days, any competent Linux techie knows about Docker/Rancher/K8s-type containers. They may not know OpenStack or S3. Personally I have used Docker but never touched OpenStack or S3.

xml/ses_glossary.xml
```diff
@@ -25,22 +25,85 @@
...
+     has no single point of failure.
+    </para>
+   </glossdef>
+  </glossentry>
+  <glossentry><glossterm>&ceph; Client</glossterm>
+   <glossdef>
+    <para>
+     The collection of &ceph; components which can access a &ceph; Storage
+     Cluster. These include the &ogw;, the Ceph Block Device, the &cephfs;, and
+     their corresponding libraries, kernel modules, and FUSEs.
```
lproven, line 87: What are FUSEs here?

lproven (changes requested):

xml/ses_glossary.xml
```diff
@@ -52,20 +115,79 @@
...
+   <glossdef>
+    <para>
+     The CRUSH data placement rule that applies to a particular pool(s).
+    </para>
+   </glossdef>
+  </glossentry>
+  <glossentry><glossterm>&dashboard;</glossterm>
+   <glossdef>
+    <para>
+     A built-in web-based &ceph; management and monitoring application to
```
tbazant, line 128:
```suggestion
     A built-in Web-based &ceph; management and monitoring application to
```

xml/ses_glossary.xml
```diff
@@ -52,20 +115,79 @@
...
+     &mon; or MON is the &ceph; monitor software.
+    </para>
+   </glossdef>
+  </glossentry>
+  <glossentry><glossterm>&mgr;</glossterm>
    <glossdef>
     <para>
-     A cluster node that maintains maps of cluster state, including the monitor
-     map, or the OSD map.
+     &mgr; or MGR is the &ceph; Manager software, which collects all the state
```
tbazant, line 167:
```suggestion
     &mgr; or MGR is the &ceph; manager software, which collects all the state
```

xml/ses_glossary.xml
```diff
@@ -52,20 +115,79 @@
...
+     from the whole cluster in one place.
+    </para>
+   </glossdef>
+  </glossentry>
+  <glossentry><glossterm>Node</glossterm>
+   <glossdef>
+    <para>
+     Any single machine or server in a Ceph cluster.
```
tbazant, line 175:
```suggestion
     Any single machine or server in a &ceph; cluster.
```

tbazant (changes requested):
minor nits, agree with Liam's notes

xml/ses_glossary.xml
asettle: You and @lproven have changed this back and forth. In theory, it should really be lowercase (which is what I had) and I will go with this as this is consistent with upstream.

xml/ses_glossary.xml
asettle: It is equally safe to say that any engineer who is familiar with docker/k8s, are familiar with OpenStack. That's not the question here. I will muse on this further, but when it comes down to it: Those looking for glossary-type clarifications in our SES docmentation should be already familiar with what this is.

I'm actually going to remove this, this hinders rather than helps. This should be a definition list. Explanations and contextual understanding goes in the product docs.

xml/ses_glossary.xml
asettle: https://docs.ceph.com/en/latest/man/8/ceph-fuse/

lproven: OK. I suspected as much. So why it is plural?

How  about "FUSE clients" instead?

xml/ses_glossary.xml
lproven: Hmmm. Either

... applies to particular pools...

or

... applies to a particular pool or pools...

?

xml/ses_glossary.xml
lproven: Aha! OK.

If I read through these in advance, before commenting, I find that 2nd time around I was expecting some things and my eye skips over them, resulting in me forgetting to comment.

That's my excuse and I'm sticking to it.

xml/ses_glossary.xml
```diff
@@ -25,20 +25,77 @@
...
      locations.
     </para>
-    <important>
-     <title>Do Not Mix with S3 Buckets</title>
-     <para>
-      S3 <emphasis>buckets</emphasis> or <emphasis>containers</emphasis>
-      represent different terms meaning <emphasis>folders</emphasis> for
-      storing objects.
-     </para>
-    </important>
```
lproven, line 41: So we're just dropping it? OK, drastic answer but an answer.

lproven (approved):

tbazant (approved):
lgtm

PR #838, opened 2020-12-16
Update manager_modules.xml with proper command to get balancer configuration

lezare:
To get the balancer configuration the correct command should be `ceph balancer status` since what is returned with `ceph config-key dump` is much more than just the balancer configuration and is confusing if one is only looking for the balancer specific information.

### DIFF
--- a/xml/manager_modules.xml
+++ b/xml/manager_modules.xml
@@ -69,7 +69,7 @@
    <para>
     To view the current balancer configuration, run:
    </para>
-<screen>&prompt.cephuser;ceph config-key dump</screen>
+<screen>&prompt.cephuser;ceph balancer status</screen>
   </tip>
 
   <important>


smithfarm (approved):

smithfarm:
@lezare Are you going to open a SES7 PR, too?

lezare:
@smithfarm I did now, see https://github.com/SUSE/doc-ses/pull/839

xml/manager_modules.xml
```diff
@@ -69,7 +69,7 @@
    <para>
     To view the current balancer configuration, run:
    </para>
-<screen>&prompt.cephuser;ceph config-key dump</screen>
+<screen>&prompt.cephuser;ceph balancer status</screen>
```
smithfarm, line 72: @lezare I wasn't expecting you to merge this so fast... next time maybe give the other reviewers a chance to look at it?

I had another idea - the command `ceph balancer status` presumably shows the balancer's status, not necessarily its configuration. I tried it on a SES6 cluster:

```
master:~ # ceph balancer status
{
    "plans": [],
    "active": false,
    "last_optimize_started": "",
    "last_optimize_duration": "",
    "optimize_result": "",
    "mode": "none"
}
```

Would it make more sense to write here:

```
Current Balancer Status

To view the current balancer status, run:

$ ceph balancer status
```

?

smithfarm: It seems misleading to say "here's how to get the current balancer _configuration_", and then give a command that displays the balancer's _status_

lezare: @smithfarm I spoke directly via chat with Alex before merging. But it does show the configuration as well, mode and plan are config values, right?

PR #839, opened 2020-12-16
Get proper balancer status output

lezare:
To get the balancer configuration the correct command should be `ceph balancer status` since what is returned with `ceph config-key dump` is much more than just the balancer configuration and is confusing if one is only looking for the balancer specific information.

### DIFF
--- a/xml/manager_modules.xml
+++ b/xml/manager_modules.xml
@@ -64,7 +64,7 @@
    <para>
     To view the current balancer configuration, run:
    </para>
-<screen>&prompt.cephuser;ceph config-key dump</screen>
+<screen>&prompt.cephuser;ceph balancer status</screen>
   </tip>
 
   <sect2 xml:id="mgr-balancer-crush-compat">


asettle (approved):

xml/manager_modules.xml
```diff
@@ -64,7 +64,7 @@
    <para>
     To view the current balancer configuration, run:
    </para>
-<screen>&prompt.cephuser;ceph config-key dump</screen>
+<screen>&prompt.cephuser;ceph balancer status</screen>
```
smithfarm, line 67: It seems misleading to say "here's how to get the current balancer _configuration_", and then give a command that displays the balancer's _status_.

Would it make more sense to write here:

```
Current Balancer Status

To view the current balancer status, run:

$ ceph balancer status
```

?

smithfarm:
I don't know... configuration and status are two very different things. 

`ceph balancer status` displays the status. But the text says the command displays the configuration. I'm confused.

xml/manager_modules.xml
lezare: But it does show the configuration as well, mode and plan are config values, right?

smithfarm: ok, how about "Current Balancer Status and Basic Configuration" ?

asettle:
Changes superceded by #841

xml/manager_modules.xml
lezare: That makes sense, I adjusted accordingly, but we now have a new PR, sorry about that.

PR #842, opened 2020-12-17
Adding instructions on removing LTSS packages

### LINKED ISSUE #715: Procedure 'Node Upgrade Using the SUSE Distribution Migration System' does not work.
**Link To Section**
https://documentation.suse.com/ses/6/html/ses-all/cha-ceph-upgrade.html#upgrade-one-node-auto

**Summary**
The procedure (6.8.2 Node Upgrade Using the SUSE Distribution Migration System ) as it is currently doesn't work for systems registered against SCC.
After step 3 the migration will start and fails with the following message:

[INFO ]: 09:11:13 | Running prepare service
[INFO ]: 09:11:13 | Bind mounting /etc/zypp
[INFO ]: 09:11:13 | Bind mounting /usr/lib/zypp/plugins
[INFO ]: 09:11:13 | Mounting kernel file systems inside /system-root
[INFO ]: 09:11:14 | Creating backup of Product data
[INFO ]: 09:11:14 | Updating Base Product to be suitable for migration
[INFO ]: 09:11:14 | Running migrate service

Executing 'zypper --root /system-root --non-interactive --gpg-auto-import-keys refresh'

Repository 'SUSE-Enterprise-Storage-5-Pool' is up to date.
Building repository 'SUSE-Enterprise-Storage-5-Pool' cache [....done]
Repository 'SUSE-Enterprise-Storage-5-Updates' is up to date.
Building repository 'SUSE-Enterprise-Storage-5-Updates' cache [....done]
Repository 'SLES12-SP3-Pool' is up to date.
Building repository 'SLES12-SP3-Pool' cache [....done]
Repository 'SLES12-SP3-Updates' is up to date.
Building repository 'SLES12-SP3-Updates' cache [....done]
All repositories have been refreshed.
Can't get available migrations from server: SUSE::Connect::ApiError: The requested products 'SUSE Linux Enterprise Server LTSS 12 SP3 x86_64' have not been activated on the system.
'/usr/lib/zypper/commands/zypper-migration' exited with status 1
[ERROR ]: 09:11:36 | migrate service failed with bash: stderr: (no output on stderr), stdout: (no output on stdout)
[INFO ]: 09:11:36 | Running grub setup service
[INFO ]: 09:11:44 | Uninstalling migration:
>>>>

**Steps To Reproduce**
This is for the scenario where a system is registered with SCC directly, so following steps 1, 2a, 3 will reproduce the error.

**Notes/Possible Solution**
The procedure is failing because the following 3 packages are installed:
SLES-LTSS
sles-ltss-release
sles-ltss-release-POOL

Adding an additional step (removing these 3 packages) to the procedure will 'fix' the problem.

asettle:
Closes #715

### DIFF
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -1109,6 +1109,28 @@ mds standby for name = mds.456-another-mds
     <link xlink:href="https://documentation.suse.com/suse-distribution-migration-system/1.0/single-html/distribution-migration-system/"/>
     to find both general and detailed information about DMS.
    </para>
+   <important>
+    <para>
+     Before starting the upgrade process, ensure the following packages are not installed:
+    </para>
+    <itemizedlist>
+     <listitem>
+      <para>
+       <package>sles-ltss-release</package>
+      </para>
+     </listitem>
+     <listitem>
+      <para>
+       <pacakge>sles-ltss-release-POOL</package>
+      </para>
+     </listitem>
+    </itemizedlist>
+    <para>
+     If the above packages are installed, you need to uninstall them with the following command
+     before proceeding.
+    </para>
+<screen>&prompt.sminion;zypper rm -y sles-ltss-release sles-ltss-release-POOL</screen>
+   </important>
    <procedure>
     <step>
      <para>
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -1109,100 +1109,113 @@ mds standby for name = mds.456-another-mds
     <link xlink:href="https://documentation.suse.com/suse-distribution-migration-system/1.0/single-html/distribution-migration-system/"/>
     to find both general and detailed information about DMS.
    </para>
-    <para>
-     Before the starting the upgrade process, check whether the <package>sles-ltss-release</package>
-     or <package>sles-ltss-release-POOL</package> packages are installed on any node
-     of the cluster:
-    </para>
-<screen>&prompt.sminion;rpm -q sles-ltss-release
-&prompt.sminion;rpm -q sles-ltss-release-POOL</screen>
-    <para>
-     If either or both are installed, remove them:
-    </para>
-<screen>&prompt.sminion;zypper rm -y sles-ltss-release sles-ltss-release-POOL</screen>
-    <important>
-      <para>
-        This must be done on all nodes of the cluster before proceeding.
-      </para>
-    </important>
-   <procedure>
-    <step>
+   <sect3 xml:id="upgrade-node-before-you-begin">
+     <title>Before You Begin</title>
      <para>
-      Install the migration RPM packages. They adjust the &grub; boot loader to
-      automatically trigger the upgrade on next reboot. Install the
-      <package>SLES15-SES-Migration</package> and
-      <package>suse-migration-sle15-activation</package> packages:
+      Before the starting the upgrade process, check whether the <package>sles-ltss-release</package>
+      or <package>sles-ltss-release-POOL</package> packages are installed on any node
+      of the cluster:
      </para>
-<screen>&prompt.sminion;zypper install SLES15-SES-Migration suse-migration-sle15-activation</screen>
-    </step>
-    <step>
-     <substeps>
+ <screen>&prompt.sminion;rpm -q sles-ltss-release
+ &prompt.sminion;rpm -q sles-ltss-release-POOL</screen>
+     <para>
+      If either or both are installed, remove them:
+     </para>
+ <screen>&prompt.sminion;zypper rm -y sles-ltss-release sles-ltss-release-POOL</screen>
+     <important>
+       <para>
+         This must be done on all nodes of the cluster before proceeding.
+       </para>
+     </important>
+     <note>
+       <para>
+         Ensure you also follow the <xref linkend="upgrade-verify-state"/> guidelines.
+         The upgrade must not be started until all nodes are fully patched. See
+         <xref linkend="upgrade-prepare-patch"/> for more information.
+       </para>
+     </note>
+   </sect3>
+   <sect3 xml:id="upgrade-nodes-dms">
+     <title>Upgrading Nodes</title>
+     <procedure>
       <step>
        <para>
-        If the node being upgraded <emphasis role="bold">is</emphasis>
-        registered with a repository staging system such as SCC, SMT, RMT, or
-        &susemgr;, create the
-        <filename>/etc/sle-migration-service.yml</filename> with the following
-        content:
+        Install the migration RPM packages. They adjust the &grub; boot loader to
+        automatically trigger the upgrade on next reboot. Install the
+        <package>SLES15-SES-Migration</package> and
+        <package>suse-migration-sle15-activation</package> packages:
        </para>
-<screen>
-use_zypper_migration: true
-preserve:
-  rules:
-    - /etc/udev/rules.d/70-persistent-net.rules
-</screen>
+  <screen>&prompt.sminion;zypper install SLES15-SES-Migration suse-migration-sle15-activation</screen>
       </step>
       <step>
-       <para>
-        If the node being upgraded is <emphasis role="bold">not</emphasis>
-        registered with a repository staging system such as SCC, SMT, RMT, or
-        &susemgr;, perform the following changes:
-       </para>
        <substeps>
         <step>
          <para>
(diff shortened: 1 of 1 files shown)


xml/admin_ceph_upgrade.xml
```diff
@@ -1109,6 +1109,28 @@ mds standby for name = mds.456-another-mds
...
+       <package>sles-ltss-release-POOL</package>
+      </para>
+     </listitem>
+    </itemizedlist>
+    <para>
+     If the above packages are installed, you can uninstall them with the following command
+     before proceeding:
+    </para>
+<screen>&prompt.sminion;zypper rm -y sles-ltss-release sles-ltss-release-POOL</screen>
+   </important>
```
asettle, line 1133: The only issue I have with this addition is we're saying "ensure these packages are installed but if they are, uninstall them" without a lot of context. Personally, I'd prefer some contextual understanding here? 
That, or perhaps we include it as part of the instruction set and add new steps:
1. Check if these packages are installed <command>
2. Optional: uninstall these packages <command>

smithfarm: Before the starting the upgrade process, check whether the `sles-ltss-release` and/or `sles-ltss-release-POOL` packages are installed:

```
# rpm -q sles-ltss-release
# rpm -q sles-ltss-release-POOL
```

If either or both are installed, remove them:

```
# zypper rm -y PACKAGE
```

asettle: I'll take that as an agreement to my second suggestion ;) I like this much better.

smithfarm: (added two important caveats to the originally proposed text)

holgis (approved):

xml/admin_ceph_upgrade.xml
```diff
@@ -1109,6 +1109,28 @@ mds standby for name = mds.456-another-mds
     <link xlink:href="https://documentation.suse.com/suse-distribution-migration-system/1.0/single-html/distribution-migration-system/"/>
     to find both general and detailed information about DMS.
    </para>
+   <important>
+    <para>
+     Before starting the upgrade process, ensure the following packages are not installed:
```
lproven, line 1114:
```suggestion
     Before starting the upgrade process, ensure the following packages are <emphasis>not</emphasis> installed:
```

lproven (changes requested):

smithfarm:
@lproven @holgis and other reviewers: My understanding is that Alex is going to re-do this along the lines of https://github.com/SUSE/doc-ses/pull/842#discussion_r545183871

So maybe hold off on the reviews until she gets to that?

xml/admin_ceph_upgrade.xml
```diff
@@ -1109,6 +1109,22 @@ mds standby for name = mds.456-another-mds
     <link xlink:href="https://documentation.suse.com/suse-distribution-migration-system/1.0/single-html/distribution-migration-system/"/>
     to find both general and detailed information about DMS.
    </para>
+    <para>
```
asettle, line 1112: I'm also tempted to put this in its own procedure with a title such as "Before you Begin" - any thoughts? I do think it is clear enough as it is, but if anyone sees that as being a good way to call it out.

smithfarm: I agree. There will definitely be multiple things to check before starting the upgrade process. For example, we will need to emphasize (as always), that the upgrade must not be started until all the SES6 nodes are fully patched.

asettle: Good point, I don't believe we do a decent job of highlighting that. I'll add the new section :)

smithfarm: (If we forgot that one, we'd be in deep trouble, because then customers on unpatched SES6 would commence the upgrade process and it would blow up, and then they'd point to the documentation and say: "nothing here says we need to patch first!")

asettle: Story of my life?

asettle:
Screenshot from 2020-12-18 11-38-32
This is how I've updated it. Added the before you begin as a sub-section. I think this looks better. I tried to play around with the callouts to see which looked better, this was my personal preference. Doing it as a procedure looked... odd.

smithfarm:
Whoa, I didn't realize that this is for the SES5 -> 6 upgrade. For that, we already have a section that says the nodes must be fully patched before starting.

I also didn't realize that there's a _separate_, 4th-level section for upgrading via the Distribution Migration System, and that that's what we're editing here.

It seems like having two separate sections with upgrade procedures is not really a good idea? Could the "upgrading via the Distribution Migration System" bits be integrated into the main upgrade procedure?

asettle:
@smithfarm yeah absolutely, admittedly why I've found this all a bit odd to begin with. But that's the beauty of reading this all from a XML POV.

Also, we should note we already also have this: https://documentation.suse.com/ses/6/single-html/ses-deployment/#upgrade-prepare-repos

asettle:
> Whoa, I didn't realize that this is for the SES5 -> 6 upgrade. For that, we already have a section that says the nodes must be fully patched before starting.

Which is why I've just cross-referenced the section, and not rewritten it.

> 
> I also didn't realize that there's a _separate_, 4th-level section for upgrading via the Distribution Migration System, and that that's what we're editing here.

Yep. 

> 
> It seems like having two separate sections with upgrade procedures is not really a good idea? Could the "upgrading via the Distribution Migration System" bits be integrated into the main upgrade procedure?

I think we're having a bit of an issue with the upgrade process in general being unclear, perhaps?

smithfarm:
@asettle Presenting various parts of the upgrade procedure in multiple sections sounds chaotic. Would it be easier for readers to understand and follow if we had a single chapter for the SES5 -> 6 upgrade procedure and organized it (i.e. the whole chapter) as a series of steps?

Moreover, I believe this bit about checking for the LTSS and getting rid of it if it's there applies to everyone doing the upgrade, not just folks doing it via some "Distribution Migration System".

asettle:
> @asettle Presenting various parts of the upgrade procedure in multiple sections sounds chaotic. Would it be easier for readers to understand and follow if we had a single chapter for the SES5 -> 6 upgrade procedure and organized it (i.e. the whole chapter) as a series of steps?

Definitely. But I'm sorry, it's not in scope to spend hours changing the way this upgrade procedure is displayed right now. 

> 
> Moreover, I believe this bit about checking for the LTSS and getting rid of it if it's there applies to everyone doing the upgrade, not just folks doing it via some "Distribution Migration System".

Quite. But should we not be combining similar/relevant information? Especially the way it is currently set out.

smithfarm:
I guess the main point I'm trying to make here is that it seems wrong to me to hide away the instruction to check for the LTSS and remove it if it's there in a subsection entitled "Node Upgrade Using the Distribution Migration System", as if this only applied to folks using some "Distribution Migration System". This instruction applies to everyone doing the upgrade.

timscotland:
> I guess the main point I'm trying to make here is that it seems wrong to me to hide away the instruction to check for the LTSS and remove it if it's there in a subsection entitled "Node Upgrade Using the Distribution Migration System", as if this only applied to folks using some "Distribution Migration System". This instruction applies to everyone doing the upgrade.

While I 100% agree that the documentation layout could be improved to make it more simple to follow, the removal of LTSS is only required in the DMS method (when DMS runs it checks the repositories and hence fails)   In the manual method you boot from ISO and that makes no reference to the existing repos (hence does not cause an issue)

smithfarm:
In that case, I suggest to:

1. structure the upgrade chapter as a series of steps
2. at the beginning of the chapter, explain that there are two methods of upgrading: DMS and ISO, what they are, what the difference between them is
3. label each step in the procedure so it's clear that the step applies to the "ISO method", "DMS method", or "both methods"

asettle:
@timscotland @smithfarm Really appreciative of the feedback you are providing. At this time, I am unable to provide suitable resources to the effort of restructuring this chapter. I propose that we include this note ASAP, closing the initial issue and covering any future customer cases, and open a new issue with Nathan's (much needed) suggested changes.

Is this a suitable compromise at this time? :)

smithfarm:
> Is this a suitable compromise at this time? :)

I'm very happy with my comment being spun off into issue #843 - thank you!

asettle:
Thanks @smithfarm ! :smile:  okay, with that in mind. Please proceed with reviews on this PR as it is, and I will look into an appropriate timeline for fixing these docs.

smithfarm (approved):

lproven (approved):

PR #852, opened 2021-01-05
cephadm user needs sudo

lezare:
By default, the cephadm user can not execute ceph commands, it is configured to execute the commands using sudo via the "/etc/sudoers.d/ceph-salt" file.


tbazant (approved):
lgtm

PR #880, opened 2021-02-02
Mention SES6 release notes instead of general speech (#863)

### LINKED ISSUE #863: [SES6] SES5 -> 6 upgrade chapter: not clear which release notes are intended to be read
There is a fundamental problem with "Read the Release Notes" subsection of the SES5 -> 6 upgrade chapter:

```
6.2 Steps to Take before Upgrading the First Node

6.2.1 Read the Release Notes

In the release notes, you can find additional information on changes since the previous release of 
SUSE Enterprise Storage. Check the release notes to see whether:

* Your hardware needs special considerations
* Any used software packages have changed significantly
* Special precautions are necessary for your installation.

The release notes also provide information that could not make it into the manual on time. They 
also contain notes about known issues. After having installed the package release-notes-ses, find 
the release notes locally in the directory /usr/share/doc/release-notes or online at 
https://www.suse.com/releasenotes/.
```

The problem, as I see it, is as follows: this subsection tells the reader to "Read the Release Notes", but it's not clear what is meant by "the release notes"! This is a section of the SES6 documentation so it would be reasonable for the reader to assume the SES6 release notes are meant.

However, this subsection is - damningly - part of a larger section entitled "Steps to Take before Upgrading the First Node". That means: when the reader is taking this step, all cluster nodes are still on SES5. The text does not make it clear, but the implication is that the reader is to read the SES6 release notes **before** upgrading the first node. Yet, if the reader follows what is written ("after having installed the package release-notes-ses...") the SES5 release notes are installed. This is clearly wrong, since the SES5 release notes say nothing about SES6 or the implications of upgrading the cluster to it.

Therefore, I propose the following changes to this section:

1. change each instance of "the release notes" to "the SES6 release notes" (for clarity)
2. delete the passage instructing the reader to install the release-notes-ses package and consult the directory /usr/share/doc/release-notes
3. inform the reader that she should read the SES6 release notes online (since this reading of the SES6 release notes is supposed to take place **before** the upgrade, it doesn't make sense to give instructions for local reading that only work **after** the upgrade of at least one node has been completed)

tbazant:
closes #863

### DIFF
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -67,9 +67,9 @@
   <sect2 xml:id="before-upgrade-release-notes">
    <title>Read the Release Notes</title>
    <para>
-    In the release notes, you can find additional information on changes since
-    the previous release of &productname;. Check the release notes to see
-    whether:
+    In the SES6 release notes, you can find additional information on changes
+    since the previous release of &productname;. Check the SES6 release notes
+    online to see whether:
    </para>
    <itemizedlist>
     <listitem>
@@ -89,13 +89,7 @@
     </listitem>
    </itemizedlist>
    <para>
-    The release notes also provide information that could not make it into the
-    manual on time. They also contain notes about known issues.
-   </para>
-   <para>
-    After having installed the package <package>release-notes-ses</package>,
-    find the release notes locally in the directory
-    <filename>/usr/share/doc/release-notes</filename> or online at
+    You can find SES6 release notes online at
     <link xlink:href="https://www.suse.com/releasenotes/"/>.
    </para>
   </sect2>
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -89,7 +89,7 @@
     </listitem>
    </itemizedlist>
    <para>
-    You can find SES6 release notes online at
+    You can find SES &productnumber; release notes online at
     <link xlink:href="https://www.suse.com/releasenotes/"/>.
    </para>
   </sect2>
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -68,7 +68,7 @@
    <title>Read the Release Notes</title>
    <para>
     In the SES &productnumber; release notes, you can find additional information on changes
-    since the previous release of &productname;. Check the SES6 release notes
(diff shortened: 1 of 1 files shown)


xml/admin_ceph_upgrade.xml
```diff
@@ -67,9 +67,9 @@
   <sect2 xml:id="before-upgrade-release-notes">
    <title>Read the Release Notes</title>
    <para>
-    In the release notes, you can find additional information on changes since
-    the previous release of &productname;. Check the release notes to see
-    whether:
+    In the SES6 release notes, you can find additional information on changes
```
tbazant, line 70:
```suggestion
    In the SES &productnumber; release notes, you can find additional information on changes
```

xml/admin_ceph_upgrade.xml
```diff
@@ -67,9 +67,9 @@
...
+    since the previous release of &productname;. Check the SES6 release notes
```
tbazant, line 71:
```suggestion
    since the previous release of &productname;. Check the SES &productnumber; release notes
```

xml/admin_ceph_upgrade.xml
```diff
@@ -89,13 +89,7 @@
...
    </itemizedlist>
    <para>
-    The release notes also provide information that could not make it into the
-    manual on time. They also contain notes about known issues.
-   </para>
-   <para>
-    After having installed the package <package>release-notes-ses</package>,
-    find the release notes locally in the directory
-    <filename>/usr/share/doc/release-notes</filename> or online at
+    You can find SES6 release notes online at
```
tbazant, line 92:
```suggestion
    You can find SES &productnumber; release notes online at
```

tbazant (commented):
Replaced fixed number with entity

lproven (approved):

smithfarm:
Seems OK to me, but I'm not very knowledgeable about the release notes. Maybe @asettle could suggest a better person to review this?

It did occur to me that it might be good (in addition to pointing to the online release notes) to also explain to the reader how to install the `release-notes-ses` RPM and access the release notes that way in case they want to read the release notes offline.

smithfarm:
One more thought - maybe this change would be good to make in the SES7 documentation as well?

smithfarm:
Ah, never mind - after re-reading #863 I see this is just fine!

(Though I still think we might need to do a similar thing to the SES7 upgrade instructions - to the extent they mention release notes)

PR #887, opened 2021-02-18
Note for mapped block devices (BSC#1176488)

asettle:


xml/deployment_hwrecommend.xml
```diff
@@ -553,6 +553,12 @@
      </para>
     </listitem>
    </itemizedlist>
+     <note>
+       <para>
+         Ensure that OSD nodes do not have any networked block devices mapped,
```
lproven, line 558: "Networked"? Would "network block devices" be enough?

Otherwise, fine!

asettle: I did think about this. The block devices themselves are being put on a network (is my understanding), as opposed to the block devices being a network on a map. 
Although I think I may have read this sentence too many times.

lproven: OK. I think just "network" is better TBH but it is not the end of the world.

lproven (approved):

PR #888, opened 2021-02-19
SES7: Add docs about newly introduced GRAFANA_FRONTEND_API_URL setting

p-se:
Add section about newly added GRAFANA_FRONTEND_API_URL to be able to
configure Grafana so that it can be reached when the Grafana users are
in a different DNS zone than the Ceph Dashboard back-end.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1176390

### DIFF
--- a/xml/admin_monitoring_alerting.xml
+++ b/xml/admin_monitoring_alerting.xml
@@ -70,9 +70,9 @@
  </para>
  <para>
   By default, traffic to &grafana; is encrypted with TLS. You can either supply
-  your own TLS certificate or use a self-signed one. A self-signed certificate
-  is automatically created and configured for &grafana; if no custom certificate
-  has been configured before &grafana; has been deployed.
+  your own TLS certificate or use a self-signed one. If no custom certificate
+  has been configured before &grafana; has been deployed, then a self-signed
+  certificate is automatically created and configured for &grafana;.
  </para>
  <para>
   Custom certificates for &grafana; can be configured using the following
--- a/xml/admin_monitoring_alerting.xml
+++ b/xml/admin_monitoring_alerting.xml
@@ -44,7 +44,7 @@
   <listitem>
    <para>
     <emphasis role="bold">&grafana;</emphasis> is the visualization and
-    alerting software. The alerting functionality of Grafana is not utilized by
+    alerting software. The alerting functionality of &grafana; is not utilized by
     this monitoring stack. For alerting, the Alertmanager is used.
    </para>
   </listitem>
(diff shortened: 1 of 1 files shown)


tbazant (approved):
LGTM

xml/admin_monitoring_alerting.xml
```diff
@@ -71,8 +71,8 @@
  <para>
   By default, traffic to &grafana; is encrypted with TLS. You can either supply
   your own TLS certificate or use a self-signed one. A self-signed certificate
-  is automatically created and configured for Grafana if no custom certificate
-  has been configured before Grafana has been deployed.
+  is automatically created and configured for &grafana; if no custom certificate
+  has been configured before &grafana; has been deployed.
```
lproven, line 75: As lines in between have been removed, I cannot make a `suggestion`. So:

This specifies  what happens before the conditions that might cause it to happen. That is not ideal.

I suggest:

«
If no custom certificate has been configured before &grafana; has been deployed,
then a self-signed certificate is automatically created and configured for &grafana;.
»

xml/admin_monitoring_alerting.xml
```diff
@@ -212,6 +212,55 @@
...
+    existence of &grafana; Dashboards before the frontend even loads them. Due to
+    the nature of how &grafana; is implemented in &dashboard;, this means that
+    two working connections are required in order to be able to see &grafana;
+    graphs in &dashboard;:
+  </para>
+
+  <itemizedlist>
+    <listitem>
+      <para>
+        The backend (Ceph Mgr module) needs to verify the existence of the
```
lproven, line 229:
```suggestion
        The backend (Ceph MGR module) needs to verify the existence of the
```

lproven: As per comment to previous PR: I think there should be an entity for Ceph...?

xml/admin_monitoring_alerting.xml
```diff
@@ -212,6 +212,55 @@
...
+        requested graph. If this request succeeds, it lets the frontend know
+        that it can safely access &grafana;.
+      </para>
+    </listitem>
+    <listitem>
+      <para>
+        The frontend then requests the &grafana; graphs directly from the user's
+        browser using an iframe. The &grafana; instance is accessed directly
```
lproven, line 237:
```suggestion
        browser using an <literal>iframe</literal>. The &grafana; instance is accessed directly
```

lproven (changes requested):

lproven (approved):

PR #889, opened 2021-02-19
Add ability to build docs from non-SUSE distros

p-se:
I know that there was a Dockerfile once, but it has been removed (not sure why, I used it [rarely]). Since then I've had to rely on these files. Furthermore, I always have to cherry pick those changes into my working branch to be able to create the documentation. As I don't use a SUSE distribution, this is the only way for me to check my work locally. Surely, `daps` is available on various distros (like Ubuntu) but it does not contain the also required styling package `suse-xsl-stylesheets`, which, by the way, also cannot be simply copied to Ubuntu to make daps work on these docs. So, in the hope that things get easier for me but will also enable every interesting party, independent of the distribution used, to check code locally before submitting a pull request, I created this pull request.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -74,3 +74,10 @@ sudo /sbin/OneClickInstallUI https://gitlab.nue.suse.com/susedoc/doc-ymp/raw/mas
 ** `$ daps -d DC-<YOUR_BOOK> html`: Build multi-page HTML document
 ** `$ daps -d DC-<YOUR_BOOK> optipng`: Always optimize new PNG images
 ** Learn more at https://opensuse.github.io/daps
+* To contribute from a non-SUSE distribution, you can use the provided
+  Dockerfile/docker-compose.yml files.
+* Using DAPS from Docker is similarly easy:
+** `$ docker-compose run daps -d DC-<YOUR_BOOK> validate`: Make sure what you have written is
+    well-formed XML and valid DocBook 5
+** `$ docker-compose run daps -d DC-<YOUR_BOOK> pdf`: Build a PDF document
+** `$ docker-compose run daps -d DC-<YOUR_BOOK> html`: Build multi-page HTML document
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,8 @@
 *.orig
 *.swp
 
+# when working with docker
+.viminfo
+.bash_history
+
 build/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM opensuse/leap
+RUN zypper -n in sudo tar daps
+RUN useradd -m user && \
+    echo "user ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
+USER user
+WORKDIR /home/user
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,17 @@
+version: '3.4'
+
+services:
+  daps:
+    build:
+      context: .
+      network: host
+    container_name: daps
+    hostname: daps
+    network_mode: host
+    tty: true
+    stdin_open: true
+    volumes:
+      - type: bind
+        source: .
+        target: /home/user
+    entrypoint: daps


sknorr:
Thanks for this -- but you could also just use https://github.com/openSUSE/daps2docker (works from the repo, or the openSUSE package) which tries to solve most of the logistics around using Docker already, making it somewhat more beginner-friendly than you personally might need.

The Dockerfile was originally to be used for CI and was never really necessary for that purpose because we just used the pre-built susedoc/doc-ci:latest container and mounted a single dir within it. That can be done just as well from a single Docker command line.

@tbazant Some repos have daps2docker instructions in them already. I'd leave it up to you whether you want to take this PR or rather copy the daps2docker instructions from e.g. https://github.com/SUSE/doc-slesforsap/#using-daps2docker .

lproven (approved):

p-se:
> Thanks for this -- but you could also just use https://github.com/openSUSE/daps2docker (works from the repo, or the openSUSE package) which tries to solve most of the logistics around using Docker already, making it somewhat more beginner-friendly than you personally might need.
> 
> The Dockerfile was originally to be used for CI and was never really necessary for that purpose because we just used the pre-built susedoc/doc-ci:latest container and mounted a single dir within it. That can be done just as well from a single Docker command line.
> 
> @tbazant Some repos have daps2docker instructions in them already. I'd leave it up to you whether you want to take this PR or rather copy the daps2docker instructions from e.g. https://github.com/SUSE/doc-slesforsap/#using-daps2docker .


This is good to know. I've tested the project and it works for me. It seems the effort to set it up (by having to clone it and then use it from that clone or create a proper symlink) is a tad bit higher than just running `docker-compose run daps -d ...` (which would also take care of building and running the container). But the latter requires `docker-compose` to be setup correctly, of course.

However, there's another one difference I noticed. The dockerfiles in this pull request build the docs in the `doc-ses` directory and they are overwritten on consecutive calls. This means that, if opened in the browser, it might be a little more comfortable to refresh instead of having to reopen.

Additionally, the `validate` command doesn't seem to be available using daps2docker, but it is for these dockerfiles (as daps is merely made available to the host system/user).

On the other hand, this may come at the cost of some usability, as it seems that daps2docker tries to be clear about its usage whereas daps might not be as clear immediately. For instance, the output of `docker-compose run daps ...` returns something like `/home/user/build/ses-admin/html/book-storage-admin/`. In this path, the directory `/home/user/` refers to the container but is mapped to the current directory, so that `build/ses-admin/html/book-storage-admin` is made available in the current directory. Which is IMHO nice, but not obvious.

I had the feeling `daps2docker` is slower than `docker-compse run daps ...`. It turned out that it indeed is. Not that it would be of significance that one is a few seconds faster than the other, but it was interesting to see that `docker-compose run daps...` is nearly twice as fast (20 seconds compared to 40 seconds) as `daps2docker` for building HTML. I don't know why, though.

```bash
user@home ~/src/doc-ses ±wip-pse-dashboard-grafana-frontend-url⚡ » time dc run daps -d DC-ses-admin html 
/home/user/images/src/odg_masters/crush-step.odg /home/user/images/src/png/crush-step.png /home/user/images/src/svg/crush-step.svg
HTML book built with REMARKS=0, DRAFT=no and META=0:
/home/user/build/ses-admin/html/book-storage-admin/
docker-compose run daps -d DC-ses-admin html  0,50s user 0,09s system 2% cpu 20,556 total
user@home ~/src/doc-ses ±wip-pse-dashboard-grafana-frontend-url⚡ » time ../daps2docker/daps2docker.sh DC-ses-admin html
Building DC file: DC-ses-admin
Building formats: html
latest: Pulling from susedoc/ci
Digest: sha256:acccf67128214896e9c20c30c7f1a0b4aab73de6628bac94dbea311897f97b89
Status: Image is up to date for susedoc/ci:latest
docker.io/susedoc/ci:latest
Container ID: f103fee51dbd29520411e49f4b8cf3d7beff7bf72d756659b0e585e0fbc7a692
Package versions in container:
  - daps 3.0.0 
  - libxslt-tools 1.1.32 
  - libxml2-tools 2.9.7 
  - xmlgraphics-fop 2.1 
  - docbook-xsl-stylesheets 1.79.2 
  - docbook5-xsl-stylesheets 1.79.2 
  - suse-xsl-stylesheets 2.0.17 
  - suse-xsl-stylesheets-sbp 2.0.17 
  - hpe-xsl-stylesheets 1.0.2.1 
  - geekodoc 1.0.2.1 
  - novdoc 1.0_1.0.2.1 
Building DC-ses-admin
daps -d /daps_temp-hPHL58VNMp/source/DC-ses-admin html  
Your output documents are:
/tmp/daps2docker-5qLtPUuB/ses-admin/html/book-storage-admin/

../daps2docker/daps2docker.sh DC-ses-admin html  0,60s user 0,63s system 3% cpu 38,474 total
user@home ~/src/doc-ses ±wip-pse-dashboard-grafana-frontend-url⚡ » 
```

While this probably hasn't made a decision easier, it might've pointed out some differences to make a more informed decision. I do understand that you might not want to have those files in this repo, though and can also simply use daps2docker.

sknorr:
> This is good to know. I've tested the project and it works for me. It seems the effort to set it up (by having to clone it and then use it from that clone or create a proper symlink) is a tad bit higher than just running `docker-compose run daps -d ...` (which would also take care of building and running the container). But the latter requires `docker-compose` to be setup correctly, of course.

You're right, having to check out the Git repo is an extra step.

daps2docker was made with the idea that you could run it even if you'd never used Docker before. You just need the `docker` installed and it will automatically start the service for you and ask for your root password if your user account is not a member of the `docker` group. Of course, some of those system interactions limit daps2docker to running on Linux systems, Mac and Windows are not currently supported.

As you said, if you already have a Docker (and Docker Compose) setup, your option is much simpler.


> However, there's another one difference I noticed. The dockerfiles in this pull request build the docs in the `doc-ses` directory and they are overwritten on consecutive calls. This means that, if opened in the browser, it might be a little more comfortable to refresh instead of having to reopen.

That was a conscious choice. DAPS is based on GNU Make and thus will take previously-generated files into account, unless run with `--force` (but even that is not a complete guarantee), meaning that if you change parameters in between runs, you may not get the right results in all cases. Since daps2docker was built as a beginner-friendly choice, it is trying to be correct and consistent rather than trying to be fast.


 > Additionally, the `validate` command doesn't seem to be available using daps2docker, but it is for these dockerfiles (as daps is merely made available to the host system/user).

You're correct, that is not specifically exposed. However, it will do a full validation before building any output.
However, you're correct, it should be able to validate on its own.


> On the other hand, this may come at the cost of some usability, as it seems that daps2docker tries to be clear about its usage whereas daps might not be as clear immediately. For instance, the output of `docker-compose run daps ...` returns something like `/home/user/build/ses-admin/html/book-storage-admin/`. In this path, the directory `/home/user/` refers to the container but is mapped to the current directory, so that `build/ses-admin/html/book-storage-admin` is made available in the current directory. Which is IMHO nice, but not obvious.
> 
> I had the feeling `daps2docker` is slower than `docker-compse run daps ...`. It turned out that it indeed is. Not that it would be of significance that one is a few seconds faster than the other, but it was interesting to see that `docker-compose run daps...` is nearly twice as fast (20 seconds compared to 40 seconds) as `daps2docker` for building HTML. I don't know why, though.

I already outlined a few reasons why it's slower, namely:

* not taking into account previous builds
* full validation happening before the build

There's a third reason why it's slower and that is the fact that it copies files around a bit much. That last bit is stemming from the fact that we ourselves are not the most proficient Docker users.

p-se:
Thank you for the explanations!

README.adoc
```diff
@@ -68,9 +68,17 @@ sudo /sbin/OneClickInstallUI https://gitlab.nue.suse.com/susedoc/doc-ymp/raw/mas
 ----
 
 * Basic DAPS usage:
+* To contribute from a a non-SUSE distribution, you can use the provided
+  Dockerfile/docker-compose.yml files.
+* Basic daps usage:
```
lproven, line 73:
```suggestion
* Basic DAPS usage:
```

lproven: Because it is an acronym.

README.adoc
```diff
@@ -68,9 +68,17 @@ sudo /sbin/OneClickInstallUI https://gitlab.nue.suse.com/susedoc/doc-ymp/raw/mas
...
 ** `$ daps -d DC-<YOUR_BOOK> validate`: Make sure what you have written is
     well-formed XML and valid DocBook 5
 ** `$ daps -d DC-<YOUR_BOOK> pdf`: Build a PDF document
 ** `$ daps -d DC-<YOUR_BOOK> html`: Build multi-page HTML document
 ** `$ daps -d DC-<YOUR_BOOK> optipng`: Always optimize new PNG images
 ** Learn more at https://opensuse.github.io/daps
+* Using daps from Docker is similarly easy:
```
lproven, line 80:
```suggestion
* Using DAPS from Docker is similarly easy:
```

lproven (changes requested):

lproven (approved):
Disclaimer: I am only looking at the English usage here, not the code/settings.

tbazant (commented):
Sorry but I have no experience with running daps under docker, i can only comment here i'm glad you're all contributing here :-)

sknorr:
@tbazant Maybe just try it out and then decide for documenting either one of the approaches (or even both) in the repo README?
Since you're a a part-time user of a non-SUSE distro anyway, it may come in handy.

I really don't want to make a decision here, partly because I guess you know the audience that may be contributing to this repo better than me. Both approaches work.

* Patrick's approach works on most OS's, but needs a working docker-compose install beforehand and an additional Dockerfile in this repo.
* daps2docker only works on Linux with systemd but is trying to handhold you through, even if you've never used Docker before.

p-se:
I created this PR in the hope it would be accepted, so that I wouldn't have to cherry pick these changes into my working branch all the time just to get the docs compiled. I also grown accustomed to the speed of this project (compared to daps2docker) as compared previously. So, I kind of saw myself keep facing a problem, which I resolved using another (but quite similar) project, which lives outside of this repository but can comfortably be called from anywhere and provides the compilation speed I've grown used to:

https://github.com/p-se/run-daps

As I've resolved my problem and it does not seem that there's much interest in this implement, I'll simply close this PR. Should this change, we can still re-open or re-create this pull request.

PR #897, opened 2021-02-26
Use 'ganesha' as example NFS namespace

tserong:
Previously the example was "nfs". I've changed it to "ganesha" for consistency with what DeepSea created in previous SES releases.

Signed-off-by: Tim Serong <tserong@suse.com>


lproven (approved):

tbazant (approved):
lgtm

PR #899, opened 2021-02-26
Add more detail to Upgrading the Object Gateway section

tserong:
Signed-off-by: Tim Serong <tserong@suse.com>

### DIFF
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -670,8 +670,9 @@ ID   CLASS  WEIGHT   TYPE NAME       STATUS  REWEIGHT  PRI-AFF
     In &productname; &productnumber;, the &ogw;s are always configured with a
     realm, which allows for multi-site (see <xref linkend="ceph-rgw-fed"/> for
     more details) in the future. If you used a single-site &ogw; configuration
-    in &productname; &prevproductnumber;, follow these steps to migrate it to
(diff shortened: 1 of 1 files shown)


lproven (approved):
It is a little more informal in tone than we usually prefer but it's fine. :-) We can fix that in postproduction...

xml/admin_ceph_upgrade.xml
```diff
@@ -670,8 +670,9 @@ ID   CLASS  WEIGHT   TYPE NAME       STATUS  REWEIGHT  PRI-AFF
     In &productname; &productnumber;, the &ogw;s are always configured with a
     realm, which allows for multi-site (see <xref linkend="ceph-rgw-fed"/> for
     more details) in the future. If you used a single-site &ogw; configuration
-    in &productname; &prevproductnumber;, follow these steps to migrate it to
-    multi-site:
+    in &productname; &prevproductnumber;, follow these steps to add a realm.
+    If you don't plan to actually use the multi-site functionality, it's fine
+    to just use <literal>default</literal> for the realm, &zgroup; and zone names.
```
tbazant, line 675:
```suggestion
    If you do not plan to actually use the multi-site functionality, it is fine
    to just use <literal>default</literal> for the realm, &zgroup; and zone names.
```

tbazant (approved):
minor tone improvement

PR #916, opened 2021-04-08
NFS Ganesha supports  v4.1 and newer only (bsc#1184009)

tbazant:
This update emphasizes this in each NFS-related section


xml/dashboard_ganesha.xml
```diff
@@ -39,8 +37,7 @@
   </mediaobject>
  </figure>
  <para>
-  To view more detailed information about an NFS export, click its table row in
-  the table.
+  To view more detailed information about an NFS export, click its table row in the table.
```
lproven, line 40: I do not think the duplication of the word "table" is needed here.

xml/dashboard_ganesha.xml
```diff
@@ -96,21 +92,18 @@
...
-     pseudo, access type, squashing, or transport protocol.
+     Specify other NFS related options, such as supported NFS protocol version, pseudo, access type, squashing, or transport protocol.
     </para>
    </step>
    <step>
     <para>
-     If you need to limit access to specific clients only, click <guimenu>Add
-     clients</guimenu> and add their IP addresses together with access type and
-     squashing options.
+     If you need to limit access to specific clients only, click <guimenu>Add clients</guimenu> and add their IP addresses together with access type and squashing options.
```
lproven, line 106: Why the removal of line-wrapping? This file now contains many very long lines, which makes viewing them in Github difficult...

lproven (changes requested):

xml/dashboard_ganesha.xml
tbazant, line 40:
```suggestion
  To view more detailed information about an NFS export, click its table row.
```

xml/dashboard_ganesha.xml
tbazant: do you think 1 sentence per line is a bed idea? To me, it has 2 advantages:
- it immediately tells you which sentences are too long
- removing or adding a sentence, or making changes inside a sentence makes the diff easier to read actually, because no auto-line wrapping occurs when sentence become shorter or longer

tbazant (commented):
removed duplicate table

PR #921, opened 2021-04-15
Use 'ceph-volume' do zap disk before OSD deployment (bsc#1171887)

tbazant:
closes bsc#1171887

### DIFF
--- a/xml/admin_install_salt.xml
+++ b/xml/admin_install_salt.xml
@@ -673,7 +673,7 @@ minion1:
       <para>
        If the disk is managed by LVM, deactivate and delete the whole LVM
        infrastructure. Refer to
-       <link xlink:href="https://www.mentation/sles-15/book_storage/data/cha_lvm.html"/>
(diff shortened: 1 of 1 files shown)


tserong (commented):
LGTM, but just in case, please double check with Boyd (when I read the bug I'm not 100% certain whether he wanted to replace the existing instructions with the `ceph-volume lvm zap` invocation, which is what you've got here, and is much easier to do :-) or whether he wanted to _add_ that command to the existing procedure)

lproven (approved):

tbazant:
> LGTM, but just in case, please double check with Boyd (when I read the bug I'm not 100% certain whether he wanted to replace the existing instructions with the `ceph-volume lvm zap` invocation, which is what you've got here, and is much easier to do :-) or whether he wanted to _add_ that command to the existing procedure)

According to https://bugzilla.suse.com/show_bug.cgi?id=1171887#c12, I hit exactly what Boyd meant :-)

tserong:
Perfect, thanks Tomáš!

PR #929, opened 2021-04-28
Info about health warning coming with the new Ceph point release

### LINKED ISSUE #928: SES6, SES7: Describe health alerts introduced by fix for CVE-2021-20288 (Unauthorized global_id reuse in cephx)
Ceph v4.2.20 (Nautilus, SES6) and v15.2.11 (Octopus, SES7) include fixes for CVE-2021-20288. As part of the fix that CVE, a new health warning was introduced, which is _on_ by default. This affects both existing users upgrading to these versions, and new users deploying for the first time. The dashboard will show the cluster in HEALTH_WARN status, and if you run `ceph status` from a shell, you will see something like this:

```
  cluster:
    id:     3fe8b35a-689f-4970-819d-0e6b11f6707c
    health: HEALTH_WARN
            mons are allowing insecure global_id reclaim

  [...]
```

This warning means that the Ceph MONs are still allowing old, unpatched clients to connect to the cluster. This ensures existing clients can still connect while the cluster is being upgraded, but warns you that there is a problem that needs to be addressed. Once the cluster and all clients are upgraded to the latest version of Ceph, you should run `ceph config set mon auth_allow_insecure_global_id_reclaim false` to disallow unpatched clients. This also goes for new cluster deployments - assuming you have no clients to start with (or will be deploying new up-to-date clients right from the start), you can run that command immediately after the cluster is deployed to ensure the new behaviour and get rid of the warning.

A complete description, including ways of silencing the warning (which is not recommended) can be found at https://docs.ceph.com/en/latest/security/CVE-2021-20288.

We need to add something like the above to both the SES6 and SES7 documentation, to make this as obvious as possible.

tbazant:
closes #928

### DIFF
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -263,7 +263,7 @@
 &prompt.smaster;salt '*' saltutil.sync_all
 &prompt.cephuser;ceph -s
 </screen>
-   <sect3 xml:id="deploy-min-cluster-disable-insecure">
+   <sect3 xml:id="upgrade-disable-insecure">
     <title>Disable insecure clients</title>
     <para>
      Since &prevcephname; v4.2.20, a new health warning was introduced that
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -263,8 +263,34 @@
 &prompt.smaster;salt '*' saltutil.sync_all
 &prompt.cephuser;ceph -s
 </screen>
+   <sect3 xml:id="deploy-min-cluster-disable-insecure">
+    <title>Disable insecure clients</title>
+    <para>
+     Since &prevcephname; v4.2.20, a new health warning was introduced that
+     informs you that insecure clients are allowed to join the cluster. This
+     warning is <emphasis>on</emphasis> by default. The &dashboard; will show
+     the cluster in the <literal>HEALTH_WARN</literal> status and verifying the
+     cluster status on the command line informs you as follows:
+    </para>
+ <screen>
+ &prompt.cephuser;ceph status
+ cluster:
+   id:     3fe8b35a-689f-4970-819d-0e6b11f6707c
+   health: HEALTH_WARN
+   mons are allowing insecure global_id reclaim
+ [...]
+ </screen>
+    <para>
+     This warning means that the &mon;s are still allowing old, unpatched
+     clients to connect to the cluster. This ensures existing clients can still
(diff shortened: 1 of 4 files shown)


lproven (approved):

ml8mr:
Should this section include a link to the [upstream CVE report](https://docs.ceph.com/en/latest/security/CVE-2021-20288/) for additional information?

tserong:
That looks good, but in addition we also need to say to run `ceph config set mon auth_allow_insecure_global_id_reclaim false` to enable the new, fixed behaviour (i.e. to force mons to not allow the old, insecure behaviour), once all clients are also updated. This command also needs to be run on newly deployed clusters too, not just in the upgrade case.

tbazant:
> ceph config set mon auth_allow_insecure_global_id_reclaim false

Added to the PR

tserong (approved):
Thanks!

tbazant:
Added Tim's snippet into deployment and upgrade chapters. @tserong can you please have a look?

tserong (approved):

smithfarm:
Is there a SES6 version of this PR somewhere?

xml/admin_ceph_upgrade.xml
```diff
@@ -263,8 +263,34 @@
 &prompt.smaster;salt '*' saltutil.sync_all
 &prompt.cephuser;ceph -s
 </screen>
+   <sect3 xml:id="upgrade-disable-insecure">
+    <title>Disable insecure clients</title>
+    <para>
+     Since &prevcephname; v4.2.20, a new health warning was introduced that
```
smithfarm, line 269: The version number should have been `v14.2.20` here?

tbazant: The question is: do we want to consolidate and disable secure clients still on SES6 before upgrading to SES7, or after the upgrade? i vote for the former...

tbazant:
> Is there a SES6 version of this PR somewhere?

Yes - look at this commit https://github.com/SUSE/doc-ses/commit/834f8c872ca28cc7a2698b093bee737100ffe1ac

xml/admin_ceph_upgrade.xml
smithfarm: @tbazant In the commit I see `v4.2.20` but that looks like a typo - seems like a `1` is missing between `v` and `4`?

tbazant: yes, correct, going to fix now......done

PR #943, opened 2021-06-07
Fix only migrate data to an EC Pool command

lezare:
The command currently is incorrect, in as such as it by default will attempt to migrate to the pool `rbd`, and since that pool in general does not exist anymore on clusters, using the command as it currently is documented will result in the following failure:

```
rbd: error opening default pool 'rbd'
Ensure that the default pool has been created or specify an alternate pool name.
```

It is thus required to in this case again specify the existing metadata pool and image as the target metadata pool/image and only specify the new data pool using the `--data-pool` option.

### DIFF
--- a/xml/admin_operating_pools.xml
+++ b/xml/admin_operating_pools.xml
@@ -1105,7 +1105,7 @@ pool4                     1.1 MiB      13      0     39                  0      
        instead:
       </para>
 <screen>
-&prompt.cephuser;rbd migration prepare <replaceable>SRC_POOL</replaceable>/<replaceable>IMAGE</replaceable> \
+&prompt.cephuser;rbd migration prepare <replaceable>SRC_POOL</replaceable>/<replaceable>IMAGE</replaceable> <replaceable>SRC_POOL</replaceable>/<replaceable>IMAGE</replaceable> \
  --data-pool <replaceable>TARGET_POOL</replaceable>/<replaceable>IMAGE</replaceable>
 </screen>
      </tip>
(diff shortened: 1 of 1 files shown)


xml/admin_operating_pools.xml
```diff
@@ -1105,7 +1105,7 @@ pool4                     1.1 MiB      13      0     39                  0
        instead:
       </para>
 <screen>
-&prompt.cephuser;rbd migration prepare <replaceable>SRC_POOL</replaceable>/<replaceable>IMAGE</replaceable> \
+&prompt.cephuser;rbd migration prepare <replaceable>SRC_POOL</replaceable>/<replaceable>IMAGE</replaceable> <replaceable>SRC_POOL</replaceable>/<replaceable>IMAGE</replaceable> \
```
lproven, line 1108: So the command now reads:

`rbd migration prepare SRC_POOL/IMAGE SRC_POOL/IMAGE --data-pool TARGET_POOL/IMAGE`

Is that right? It looks redundant to me. Why is the source present twice?

lproven: @lezare  --- any comment?

lezare: The source is present twice since the image meta data is not to be moved, that remains on the original pool, in this scenario ONLY the data are to be moved to a new EC pool. Because the meta data remains on the original replicated pool, that needs to be specified as the source AND target. If you do not do this, the command errors out and does not work.

lezare: okay, I have updated the commit as per the above suggestion.

tbazant (approved):
LGTM

PR #945, opened 2021-06-25
Add missing -i option

lezare:
The current management user commands mentions to simply specify the [PASSWORD] for the user, this will error out with for example:

`Error EINVAL: Please specify the file containing the password/secret with "-i" option.`

It is required to specify a password file containing the password using the -i option.

### DIFF
--- a/xml/dashboard_user_cmdline.xml
+++ b/xml/dashboard_user_cmdline.xml
@@ -100,8 +100,8 @@
   </para>
 
   <para>
-   User accounts are stored in &mon;’s configuration database and are globally
-   shared across all &mgr; instances.
+   User accounts are stored in &mon;'s configuration database and are shared
+   globally across all &mgr; instances.
   </para>
 
   <para>
--- a/xml/dashboard_user_cmdline.xml
+++ b/xml/dashboard_user_cmdline.xml
@@ -121,7 +121,7 @@
     <term>Create a new user:</term>
     <listitem>
 <screen>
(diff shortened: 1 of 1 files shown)


xml/dashboard_user_cmdline.xml
```diff
@@ -100,8 +100,8 @@
   </para>
 
   <para>
-   User accounts are stored in &mon;’s configuration database and are
-   globally shared across all &mgr; instances.
+   User accounts are stored in &mon;’s configuration database and are globally
+   shared across all &mgr; instances.
```
lproven, line 104:
```suggestion
   User accounts are stored in &mon;'s configuration database and are shared
   globally across all &mgr; instances.
```

lproven: Be very careful about accidentally inserting curly ‘single’ and “double” quotation marks. Our rendering and production software inserts these automatically, but it expects 'straight' quotes, like those, and not curly ones, \`backticks\` or isolated acute accents: ´

lproven (changes requested):

xml/dashboard_user_cmdline.xml
lezare: I made no changes to that part of the documentation with my original commit here? Or what am I missing?

lproven: I had to check, but you are right. Sorry for the noise. 

For clarity: are we talking about one single MON node or several MON nodes?

PR #946, opened 2021-06-25
Fix xml:id/linkend handling

tomschr:
To make it SEO compatible, replace any "-" and other characters with underscore

### DIFF
--- a/convert-manpages.sh
+++ b/convert-manpages.sh
@@ -149,7 +149,7 @@ function fix-manpage() {
     local DIR="${1}"
     local OUTDIR="${2}"
     local OUTFILE="$OUTDIR/$BASE"
-    local PARAMS="--stringparam id ${BASE%*.xml}-"
+    local PARAMS="--stringparam id ${BASE%*.xml}_"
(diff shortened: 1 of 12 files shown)


tbazant (approved):
what i understand looks good, the rest is pure belief :-)

tomschr:
Haha, okay. I've tested it with:

```
$ ./convert-manpages.sh -v manpages
```

That will clone the sources from an upstream repository, creates a temporary directory (e.g. `/tmp/nfs-ganesha-01yO2/`) and the result is saved in `manpages`. You can reuse the temp directory with the `-g` option, if you need to call the script again.

The GH Action still fails as I haven't committed the updated manpages into `xml`. I leave that up to you to check if it's safe to do so. :wink: Just ask if you need help.

Merge the PR whenever you think it's right.

tbazant (changes requested):
the script replaces with underscores, but the point was to remove underscores from ID's and xrefs completely :-)

tbazant (approved):
LGTM

PR #948, opened 2021-08-05
Pool name should be specified first

lezare:
Command syntax is incorrect for marking an EC pool as an RBD pool, the pool name needs to be specified before the application (we do seem to have this properly set elsewhere in the documentation), else (with the command as it is now in this part of the documentation) the following error will be returned since we then look for a pool name called `rbd` (which does by default not exist):

`Error ENOENT: unrecognized pool 'rbd'`


lproven (approved):

tbazant (approved):
LGTM

PR #953, opened 2021-09-06
Fixed syntax of disks.deploy command (bsc#1189781)

tbazant:


ebl-nde (commented):
If you replaced the example `target=df0-4d-a2-75-6f-22` with `target=data1.ceph` so it matches the previous example it would make it clearer. The MAC address is from a SOC + SES environment where the hostnames are managed by crowbar.

PR #957, opened 2021-10-05
Aligned wrong commands in Rook deployment process (bsc#1191153)

tbazant:


lproven (approved):
I can't verify the technical content, which looks very different, but otherwise: LGTM.

PR #961, opened 2021-10-20
Decreasing number of PGs is possible with Nautilus

### LINKED ISSUE #956: 35.4 'Too Many PGs per OSD' Status Message - PG merging is supported with Nautilus
**Link To Section**
https://documentation.suse.com/ses/6/single-html/ses-admin/index.html#storage-bp-recover-toomanypgs

**Summary**
The second paragraph in this section still mentions that reducing PGs for an existing pool is not supported. This is incorrect, from Nautilus (SES 6) onwards reducing of PGs is supported.

**Notes**
Suggest to either completely remove the second paragraph in this section that mentions reducing of PGs is not possible after pool creation OR adjust the second paragraph in this part of the documentation, for example to something like the following:

```
As of the Nautilus release the number of PGs can also be reduced after a pool is already created. Using <command>ceph osd pool set <replaceable>POOL_NAME</replaceable> <replaceable>PG_NUM</replaceable></command> the amount of PGs for an existing Pool can be either increased or decreased.

This is however a resource intensive operation and changes should be made incrementally, especially when decreasing (merging) the amount of PGs. It is recommended to instead have the PG autoscaler manager module enabled to manage the amount of PGs per pool. For details see <xref linkend="op-pgs-autoscaler"/>.
```

tbazant:
closes #956

### DIFF
--- a/xml/admin_ceph_troubleshooting.xml
+++ b/xml/admin_ceph_troubleshooting.xml
@@ -118,12 +118,15 @@
    is already created. To increase or decrease the amount of PGs for an
    existing pool, run the following command:
   </para>
+
 <screen>&prompt.cephuser;ceph osd pool set <replaceable>POOL_NAME</replaceable> <replaceable>NUM_OF_PG</replaceable></screen>
+
   <para>
    Note that this operation is resource intensive and you are encouraged to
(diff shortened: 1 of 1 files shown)


lezare:
LGTM.

PR #962, opened 2021-10-27
SEO: remove editurls_l10n_SES7

jfaltenbacher:
dm:editurls tag removed for all languages of SES 7.

Will be done for each branch version separately therefore no backports needed

### DIFF
--- a/l10n/es-es/xml/bp_troubleshooting_pgs.xml
+++ b/l10n/es-es/xml/bp_troubleshooting_pgs.xml
@@ -12,9 +12,9 @@
 
   <para>
    As previously noted, a placement group is not necessarily problematic
-   because its state is not <literal>active+clean</literal>. Generally,
-   Ceph's ability to self-repair may not be working when placement groups get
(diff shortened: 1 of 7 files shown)


sknorr:
@jfaltenbacher I have finally gotten to checking out the issue you told us about last week that with the CI in this PR.

Soo, as Christoph had already guessed, the issue appeared now because the GeekoDoc schema used for CI had changed since the last CI run here:

* Jul 21: Previous CI run on the maintenance/ses7 branch, commit 62800b0a7dcd51e52dbf6cb04b3b7bf07e223ff2
* Aug 08: Toms releases the updated version of GeekoDoc v1, alongside a beta of GeekoDoc v2
* Oct 27: This PR comes along.

I've also checked with the old package and sure enough, it validates on old commit. So, nothing to see here except the cost of doing backward-incompatible but (imo small & useful) changes.

jfaltenbacher:
ok, thanks for clarifying Stefan

PR #967, opened 2021-11-08
Update Dashboard SSO information

lezare:
The cert files should be present on the active MGR host, this information is specified since August in the online Ceph documentation (see https://docs.ceph.com/en/latest/mgr/dashboard/#enabling-single-sign-on-sso) but are not reflected currently in our documentation.

### DIFF
--- a/xml/dashboard_manual_config.xml
+++ b/xml/dashboard_manual_config.xml
@@ -418,7 +418,8 @@
     <listitem>
      <para>
       Optional. File path or content of the certificate that will be used by
-      &dashboard; (Service Provider) for signing and encryption (these file paths should be accessible from the active ceph-mgr instance).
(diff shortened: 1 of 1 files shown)


xml/dashboard_manual_config.xml
```diff
@@ -418,7 +418,8 @@
     <listitem>
      <para>
       Optional. File path or content of the certificate that will be used by
-      &dashboard; (Service Provider) for signing and encryption.
+      &dashboard; (Service Provider) for signing and encryption (these file
+      paths should be accessible from the active ceph-mgr instance).
```
tbazant, line 422:
```suggestion
      &dashboard; (Service Provider) for signing and encryption. These file
      paths need to be accessible from the active &mgr; instance.
```

tbazant (approved):
small change for wording but LGTM anyways

PR #971, opened 2021-11-29
Revert update of Node-Exporter from 1.0.1 to 0.18.1

p-se:
Revert update to Node-Exporter container. The issue that was tried to be solved by an update turned out to have been a mistake in merging. That way the correct version was replaced by an older one that was no longer available on registry.suse.com. This has been restored in https://github.com/SUSE/ceph/pull/445 to refer to Node-Exporter version 0.18.1. Naturally, the docs need to point to that version as well. Sorry for the fuss!

### DIFF
--- a/xml/admin_monitoring_alerting.xml
+++ b/xml/admin_monitoring_alerting.xml
@@ -163,7 +163,7 @@
    </listitem>
    <listitem>
     <para>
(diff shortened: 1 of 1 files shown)


tbazant (approved):
I hope you're right this time

PR #973, opened 2021-12-06
Fixed PTF process (bsc#1193293)

tbazant:
closes bsc#1193293

rendered PDF is attached 
[bp-troubleshooting-cephadm_color_en.pdf](https://github.com/SUSE/doc-ses/files/7668494/bp-troubleshooting-cephadm_color_en.pdf)

### DIFF
--- a/xml/bp_troubleshooting_cephadm.xml
+++ b/xml/bp_troubleshooting_cephadm.xml
@@ -406,18 +406,59 @@ cat config-json.json
   </para>
 
   <para>
-   With &cephadm;, deploying the PTF is already possible by redeploying a
-   specific daemon with a specific <literal>container_image</literal>.
+   PTFs are published on the <literal>registry.suse.com</literal> server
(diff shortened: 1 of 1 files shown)


asettle (approved):
Thanks!

lezare (commented):
Since a PTF may be related to other services, should we not also add a note on how to get the relevant mon key for the service in question? I am referring to this part from the bug description:

Determine the mon key corresponding to the desired service(s):
    # ceph auth ls | grep [DAEMON_TYPE]

tbazant:
> Since a PTF may be related to other services, should we not also add a note on how to get the relevant mon key for the service in question? I am referring to this part from the bug description:
> 
> Determine the mon key corresponding to the desired service(s): # ceph auth ls | grep [DAEMON_TYPE]

@lezare One procedure step added, please check

lezare (approved):
LGTM.

Issue #981, opened 2022-02-15
[doc] 10.2  Upgrading the Salt Master - registry authentication not a list!


Martin-Weiss:
### Type of report
   - [ x] bug
   - [ ] feature request

### Source Document
10.2  Upgrading the Salt Master:

https://documentation.suse.com/ses/7/single-html/ses-deployment/#upgrade-salt-master

### Summary
This is wrong:

```
ses7_container_image: 192.168.121.1:5000/my/ceph/image
ses7_container_registries:
  - location: 192.168.121.1:5000
ses7_container_registry_auth:
  - registry: 192.168.121.1:5000
  - username: USER_NAME
  - password: PASSWORD
```

When specifying the ses7_container_registry_auth as "lists" with "-" - the error during apply is this:

```
salt <node> state.apply ceph.upgrade.ses7.adopt
 
output:
 
<node>:
    Data failed to compile:
----------
    Rendering SLS 'base:ceph.upgrade.ses7.adopt' failed: Jinja variable 'list object' has no attribute 'get'
ERROR: Minions returned with non-zero exit code

```

### Steps To Reproduce
Follow the documentation for specification of registry authentication and the run the adopt state.

### Expected Results
Works without an error.

### Actual Results
Error.

### Notes
Please remove the "-" for the auth:

```
ses7_container_image: 192.168.121.1:5000/my/ceph/image
ses7_container_registries:
  - location: 192.168.121.1:5000
ses7_container_registry_auth:
  registry: 192.168.121.1:5000
  username: USER_NAME
  password: PASSWORD
```

See also https://github.com/SUSE/doc-ses/issues/952

Martin-Weiss:
Any idea when the official docs will be updated? It seems that we still have the "bug", there..

PR #982, opened 2022-02-15
Fixes issue #981 - 10.2 Upgrading the Salt Master - registry authenti…

Martin-Weiss:
…cation not a list!

Need to remove the "-" as the registry/username/password in ses7_container_registry_auth is not a list.


smithfarm (approved):

tbazant (approved):
LGTM

PR #991, opened 2022-03-07
Adapt Prometheus Alertmanager SNMP documentation to new upstream solution

votdev:
Adapt Prometheus Alertmanager SNMP Gateway setup to new upstream solution in SES7.1 (Pacific). This should give us less support effort.

@tbazant Please feel free to cherry-pick this PR/commit to add your changes in a separate PR.

Maybe it is possible to add those screenshots to the documentation, too.
snmp_v2
snmp_v3

### DIFF
--- a/xml/admin_monitoring_alerting.xml
+++ b/xml/admin_monitoring_alerting.xml
@@ -477,12 +477,12 @@
   </para>
  </sect1>
  <sect1 xml:id="prometheus-webhook-snmp">
-  <title>&prometheus; &alertmanager; SNMP Gateway</title>
+  <title>&prometheus; &alertmanager; SNMP gateway</title>
 
   <para>
    If you want to get notified about &prometheus; alerts via SNMP traps, then
-   you can install the &prometheus; &alertmanager; SNMP Gateway via &cephadm;
-   or the Ceph Dashboard. To do so for SNMPv2c for example, you need to create
+   you can install the &prometheus; &alertmanager; SNMP gateway via &cephadm;
+   or the &dashboard;. To do so for SNMPv2c, for example, you need to create
    a service and placement specification file with the following content:
   </para>
 
@@ -506,7 +506,7 @@ spec:
 </screen>
 
   <para>
-   Alternatively you can use the Ceph Dashboard to deploy the SNMP Gateway
+   Alternatively, you can use the &dashboard; to deploy the SNMP gateway
    service for SNMPv2c and SNMPv3.
   </para>
  </sect1>
--- a/xml/admin_monitoring_alerting.xml
+++ b/xml/admin_monitoring_alerting.xml
@@ -507,7 +507,8 @@ spec:
 
   <para>
    Alternatively, you can use the &dashboard; to deploy the SNMP gateway
-   service for SNMPv2c and SNMPv3. For more details, refer to <xref linkend="dashboard-cluster-services"/>.
+   service for SNMPv2c and SNMPv3. For more details, refer to
+   <xref linkend="dashboard-cluster-services"/>.
   </para>
  </sect1>
 </chapter>
--- a/xml/admin_monitoring_alerting.xml
+++ b/xml/admin_monitoring_alerting.xml
@@ -482,8 +482,8 @@
   <para>
    If you want to get notified about &prometheus; alerts via SNMP traps, then
    you can install the &prometheus; &alertmanager; SNMP gateway via &cephadm;
-   or the &dashboard;. To do so for SNMPv2c, for example, you need to create
-   a service and placement specification file with the following content:
+   or the &dashboard;. To do so for SNMPv2c, for example, you need to create a
+   service and placement specification file with the following content:
   </para>
 
   <note>
--- a/xml/admin_monitoring_alerting.xml
+++ b/xml/admin_monitoring_alerting.xml
@@ -482,8 +482,8 @@
   <para>
    If you want to get notified about &prometheus; alerts via SNMP traps, then
    you can install the &prometheus; &alertmanager; SNMP Gateway via &cephadm;
-   or the Ceph Dashboard. To do so for SNMPv2c for example, you need to create a service and
-   placement specification file with the following content:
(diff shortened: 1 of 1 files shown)


xml/admin_monitoring_alerting.xml
```diff
@@ -494,129 +494,20 @@
...
 
 <screen>
-service_type: container
-service_id: prometheus-webhook-snmp
+service_type: snmp-gateway
+service_name: snmp-gateway
 placement:
     <replaceable>ADD_PLACEMENT_HERE</replaceable>
-image: registry.suse.com/ses/7/prometheus-webhook-snmp:latest
-args:
```
ml8mr, line 502: Is the prometheus-webhook-snmp image no longer required (or provided someplace else)?

votdev: > Is the prometheus-webhook-snmp image no longer required (or provided someplace else)?

It will no longer be needed if we implement the following suggestion.

Ceph upstream has changed the OIDs of all [Prometheus alerts](https://github.com/ceph/ceph/commit/cfc4d3ec67292cc79e28d64582751e10fe5710af#diff-cf1112e777b119958d89b5464bda47ea828a2cb009729b0693d0d35cc0ec55c1). Because of that our SNMP webhook will not work correctly anymore without adaption or adaption of service spec in the documentation. To reduce the maintenance and prevent problems i would suggest to drop our own solution and advise our customers to switch to the official Ceph upstream SNMP gateway. This should be simple by deploying the new snmp-gateway service and delete the old one. This way the MIB and Prometheus alerts are always in sync.

We can adapt our own SNMP webhook, but this will require additional work and we always need to take care what upstream happens. Because of that i suggest to fully switch to the upstream solution; incl. customers that are using our own solution in SES7.

tbazant:
> Adapt Prometheus Alertmanager SNMP Gateway setup to new upstream solution in SES7.1 (Pacific). This should give us less support effort.
> 
> @tbazant Please feel free to cherry-pick this PR/commit to add your changes in a separate PR.
> 
> Maybe it is possible to add those screenshots to the documentation, too. snmp_v2 snmp_v3

@votdev the changes you suggest make sense to me as a whole. if you think that screenshots would help, tell me which one to insert and where, because i don't understand much the topic :-)

xml/admin_monitoring_alerting.xml
```diff
@@ -477,13 +477,13 @@
   </para>
  </sect1>
  <sect1 xml:id="prometheus-webhook-snmp">
-  <title>&prometheus; &alertmanager; SNMP webhook</title>
+  <title>&prometheus; &alertmanager; SNMP Gateway</title>
```
tbazant, line 480:
```suggestion
  <title>&prometheus; &alertmanager; SNMP gateway</title>
```

xml/admin_monitoring_alerting.xml
```diff
@@ -494,129 +494,20 @@
...
+  credentials:
+    snmp_community: <replaceable>ADD_COMMUNITY_STRING_HERE</replaceable>
+  snmp_destination: <replaceable>ADD_FQDN_HERE</replaceable>:<replaceable>ADD_PORT_HERE</replaceable>
+  snmp_version: V2c
 </screen>
 
   <para>
-   Use this service specification to get the service running using its default
-   settings.
+   Alternatively you can use the Ceph Dashboard to deploy the SNMP Gateway
```
tbazant, line 509:
```suggestion
   Alternatively, you can use the &dashboard; to deploy the SNMP gateway
```

xml/admin_monitoring_alerting.xml
```diff
@@ -477,13 +477,13 @@
...
-  <title>&prometheus; &alertmanager; SNMP webhook</title>
+  <title>&prometheus; &alertmanager; SNMP Gateway</title>
 
   <para>
    If you want to get notified about &prometheus; alerts via SNMP traps, then
-   you can install the &prometheus; &alertmanager; SNMP webhook via &cephadm;.
-   To do so, you need to create a service and placement specification file with
-   the following content:
+   you can install the &prometheus; &alertmanager; SNMP Gateway via &cephadm;
+   or the Ceph Dashboard. To do so for SNMPv2c for example, you need to create
```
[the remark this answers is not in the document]
dariavladykina:
```suggestion
   or the Ceph Dashboard. To do so for SNMPv2c, for example, you need to create
```

xml/admin_monitoring_alerting.xml
mgfritch: this will deploy the upstream `docker.io/maxwo/snmp-notifier:v1.2.1` container image via cephadm.

We should probably create a downstream build of this image and update the cephadm image paths....

asettle (approved):

xml/admin_monitoring_alerting.xml
```diff
@@ -494,129 +494,20 @@
...
+    snmp_community: <replaceable>ADD_COMMUNITY_STRING_HERE</replaceable>
+  snmp_destination: <replaceable>ADD_FQDN_HERE</replaceable>:<replaceable>ADD_PORT_HERE</replaceable>
+  snmp_version: V2c
 </screen>
 
   <para>
-   Use this service specification to get the service running using its default
-   settings.
+   Alternatively, you can use the &dashboard; to deploy the SNMP gateway
+   service for SNMPv2c and SNMPv3.
```
tbazant, line 510:
```suggestion
   service for SNMPv2c and SNMPv3. For more details, refer to <xref linkend="dashboard-cluster-services"/>.
```

tbazant (approved):
LGTM now

xml/admin_monitoring_alerting.xml
ml8mr: @votdev Any progress or thoughts on a downstream snmp-notifier (as Mike suggested)?

s0nea: Well... the problem is not only about the container. Unfortunately we need a package of the upstream repository which is not available yet.

PR #1001, opened 2022-03-21
Update container image paths to point to `ceph` instead of `prometheus`

p-se:

### DIFF
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -357,17 +357,17 @@
      </listitem>
      <listitem>
       <para>
-       registry.suse.com/ses/7/prometheus/prometheus-server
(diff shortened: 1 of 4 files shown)


p-se:
As updating just the path to the container images is not possible, because the container images would be built with the newest available version, the update of both, the container image paths for Ceph as well as the docs update are postponed until updates to the container images are released which actually use the new container images.

tbazant (commented):
LGTM

p-se:
Files already [up-to-date](https://github.com/SUSE/ceph/commit/dfe2b2888bd914ae433099edffc41da0044fe305).

tbazant (approved):
LG

ml8mr (approved):
LGTM

Issue #1004, opened 2022-03-22
[doc] 10.3 Upgrading the MON, MGR, and OSD nodes - add bluestore_fsck_quick_fix_on_mount information


tserong:
When upgrading from Ceph Nautilus to Pacific, there's a format conversion necessary for OSDs, to improve the accounting for "omap" data. This does not happen automatically. There's a few ways to make it happen:

1. Run `ceph config set osd bluestore_fsck_quick_fix_on_mount true` _before_ adopting any OSDs. As all the OSDs are adopted and started, they will do the format conversion automatically.
2. Run `ceph config set osd bluestore_fsck_quick_fix_on_mount true` _after_ adopting all the OSDs, then restart the OSDs later (they'll do the conversion on startup).
3. Don't set that option, and instead go through an irritating manual procedure using `ceph-bluestore-tool` for each OSD. For each individual OSD, this would be something like the following:

```
# cephadm unit --name osd.1 stop
# cephadm shell --mount /var/lib/ceph/f8645ef8-a3bf-46ce-9f67-6514afd211c1:/var/lib/ceph
[ceph: root@node1 /]# ceph-bluestore-tool quick-fix --path /var/lib/ceph/osd.1
(...some command output here...)
[ceph: root@node1 /]# exit
# cephadm unit --name osd.1 start
```

We need to mention how to do this conversion in the documentation, because if you _don't_ do it, `ceph status` will tell you your cluster is in HEALTH_WARN state, and `ceph health detail` will say something like this for each OSD:

```
HEALTH_WARN on 2 OSD(s)
BLUESTORE_NO_PER_POOL_OMAP on 2 OSD(s)
    osd.1 legacy (not per-pool) omap detected, suggest to run store repair to benefit from per-pool omap usage statistics
    osd.5 legacy (not per-pool) omap detected, suggest to run store repair to benefit from per-pool omap usage statistics
```

Note also that this conversion "may take a few minutes to as much as a few hours (for an HDD with lots of omap data)" (see point 4 under https://docs.ceph.com/en/latest/releases/pacific/#upgrading-non-cephadm-clusters).

So, the question is: how best do we describe the issue, and which procedure should we recommend?

- Options 1 and 2 are both a lot easier to _do_ than option 3.
- Option 1 may mean that upgrading each storage node is very slow (if it has to wait a long time to do the conversion as part of the upgrade), but it means that once all the OSDs are adopted, the cluster will be HEALTH_OK, not HEALTH_WARN
[...]

PR #1005, opened 2022-03-22
Added 1 step to OSD 6 > 7.1 upgrade procedure for format conversion

tbazant:
closes #1003 

PDF is attached
[upgrade-mon-mgr-nodes_color_en.pdf](https://github.com/SUSE/doc-ses/files/8322857/upgrade-mon-mgr-nodes_color_en.pdf)

### DIFF
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -502,6 +502,24 @@ Nodes running older software versions must be upgraded in the following order:
   <procedure>
    <step>
     <para>
+     Before adopting any OSD node, you need to perform a format
+     conversion of OSD nodes to improve the accounting for OMAP data.
+     You can do so by running the following command on the &adm;:
+    </para>
+<screen>&prompt.cephuser;ceph config set osd bluestore_fsck_quick_fix_on_mount true</screen>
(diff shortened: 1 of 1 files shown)


tserong (approved):
Thanks, I think this is by far the most straightforward way to do this procedure given the options I outlined in #1004 :-)

@mgfritch, what do you think? Do we need to add any information about those other possible ways of doing this thing?

asettle (commented):
One nit

mgfritch (approved):
> @mgfritch, what do you think? Do we need to add any information about those other possible ways of doing this thing?

this seems like the best approach, it anything maybe a note to perform the adoption on a per OSD or per node basis ... ?

tserong:
> this seems like the best approach, it anything maybe a note to perform the adoption on a per OSD or per node basis ... ?

The docs already say at the start to "Upgrade the Ceph Monitor, Ceph Manager, and OSD nodes one at a time", so I think that covers the per-node basis. We don't actually have a means for DeepSea to do the adoption per OSD, for that you'd have to run `cephadm adopt` by hand.

dariavladykina (approved):
LGTM!

PR #1026, opened 2022-05-24
Reported wrong RBD image size

### LINKED ISSUE #1024: [doc] Issue in "Creating RBD images"
### Type of report
   - [ ] bug
   - [ ] feature request

### Source Document
9.1.4.3 Exporting RBD images via iSCSI 

https://documentation.suse.com/ses/7/html/ses-all/deploy-additional.html#ceph-iscsi-rbd-export

### Summary

Need to add a note about disabling "emulate legacy capacity" functionality.

https://bugzilla.suse.com/show_bug.cgi?id=1199090

### Notes

In older kernels the rbd target incorrectly reported the image size one block (512B) larger, and the LUN could write one block out of the image size. In newer versions that issue has been fixed but silently changing the behavior for existing images could lead to data corruption due to the previously used last block not accessible any more. For this reason the new kernels by default still supports the old behavior unless emulate_legacy_capacity attribute is set to 0 for a target.

This can be done in gwcli:
```
cd /disks/{pool}/{image}
reconfigure emulate_legacy_capacity 0
```
Or it can be done in dashboard when creating or editing a target (in the target image config panel).

We recommend to disable emulate_legacy_capacity for all newly created targets. And this is required if one wants to use rbd images with object-map feature enabled.

For already existing targets, which does not have object-map feature, you may continue to use emulate_legacy_capacity. In this case you may expect to see "target_core_rbd: beyond EOD" kernel messages on the gateway nodes, which do not affect the functionality. If you want to disable emulate_legacy_capacity for existing targets, to avoid potential data damage, it is recommended to increase the image (metadata) size by on 512B to include out of the boundary block.

To see the current (reported) image size in bytes one can use:
```
rbd info ${pool}/${image} --format json --pretty-format
```
And the resize command can be used to set then new size (old_size + 512):
```
rbd resize ${pool}/${image} -s ${new_size}
```
The whole procedure could be done with this one liner:
```
size=$(rbd info ${pool}/${image} --format json | jq '.size') && test -n $size && rbd resize ${pool}/${image} -s $((size + 512))B && echo OK
```
This should be done when the image is not used (there are no open sessions for the target), just before setting emulate_legacy_capacity=0 attribute.

tbazant:
fixes #1024

### DIFF
--- a/xml/deploy_additional.xml
+++ b/xml/deploy_additional.xml
@@ -305,11 +305,9 @@
      <para>
       A bug in older kernels caused that an RBD target reported an image size
       one block (512B) larger than it actually was. Therefore, the LUN could
-      write one additional block of the image. Although this bug is fixed now,
-      existing images can be corrupted because they previously used the block
-      that is not accessible any more. For this reason, the new kernels support
-      the old behavior by default unless you set
-      <option>emulate_legacy_capacity</option> attribute to 0 for a target.
+      write one additional block of the image. To keep the image data
+      consistent, new kernels support the old behavior by default unless you
+      set <option>emulate_legacy_capacity</option> attribute to 0 for a target.
      </para>
      <para>
       We recommend disabling the old behavior for newly created targets. If you
--- a/xml/deploy_additional.xml
+++ b/xml/deploy_additional.xml
@@ -304,8 +304,8 @@
      <title>Incorrectly reported image sizes and the <option>emulate_legacy_capacity</option> option</title>
      <para>
       A bug in older kernels caused that an RBD target reported an image size
-      one block (512 bytes) larger than it actually was. Therefore, the LUN could
-      write one additional block of the image. To keep the image data
+      one block (512 bytes) larger than it actually was. Therefore, the LUN
+      could write one additional block of the image. To keep the image data
       consistent, new kernels support the old behavior by default unless you
       set <option>emulate_legacy_capacity</option> attribute to 0 for a target.
      </para>
--- a/xml/deploy_additional.xml
+++ b/xml/deploy_additional.xml
@@ -301,7 +301,7 @@
      </itemizedlist>
     </note>
     <important>
-     <title>Incorrectly reported image sizes</title>
+     <title>Incorrectly reported image sizes and the <option>emulate_legacy_capacity</option> option</title>
      <para>
       A bug in older kernels caused that an RBD target reported an image size
       one block (512B) larger than it actually was. Therefore, the LUN could
--- a/xml/deploy_additional.xml
+++ b/xml/deploy_additional.xml
@@ -300,6 +300,55 @@
       </listitem>
      </itemizedlist>
     </note>
+    <important>
+     <title>Incorrectly reported image sizes</title>
+     <para>
+      A bug in older kernels caused that an RBD target reported an image size
+      one block (512B) larger than it actually was. Therefore, the LUN could
+      write one additional block of the image. Although this bug is fixed now,
+      existing images can be corrupted because they previously used the block
+      that is not accessible any more. For this reason, the new kernels support
+      the old behavior by default unless you set
+      <option>emulate_legacy_capacity</option> attribute to 0 for a target.
+     </para>
+     <para>
+      We recommend disabling the old behavior for newly created targets. If
+      you plan to enable the <option>object-map</option> feature for RBD
+      images, disabling <option>emulate_legacy_capacity</option> is required.
+      Use the &gwcli; shell to disable it:
+     </para>
+<screen>
+&prompt.root;&gwcli;
+&prompt.gwcli;cd /disks/<replaceable>POOL</replaceable>/<replaceable>IMAGE</replaceable>
+&prompt.gwcli;reconfigure emulate_legacy_capacity 0
+</screen>
+     <tip>
+      <para>
+       You can also disable the emulation in the &dashboard; when creating or
+       editing the RBD target.
+      </para>
+     </tip>
+     <para>
+      For already existing RBD targets that do not use the
+      <option>object-map</option> feature, you may continue to use the
+      <option>emulate_legacy_capacity</option> enabled. In such case you can
+      safely ignore related <literal>target_core_rbd: beyond EOD</literal>
+      kernel messages on corresponding gateway nodes.
+     </para>
+     <para>
+      If you want to disable <option>emulate_legacy_capacity</option> for
+      existing targets to avoid potential data corruption, we recommend
+      increasing the image size by on 512B to include the additional boundary
+      block. Resize the image when it is not used right before disabling the
+      emulation. To see the current image size in bytes, use the following
+      command:
+     </para>
+<screen>&prompt.cephuser;rbd info <replaceable>POOL</replaceable>/<replaceable>IMAGE</replaceable> --format json --pretty-format</screen>
+     <para>
+      To resize the image, use the following command:
+     </para>
+<screen>&prompt.cephuser;rbd resize <replaceable>POOL</replaceable>/<replaceable>IMAGE</replaceable> -s <replaceable>REPORTED_SIZE + 512</replaceable></screen>
+</important>
     <para>
      As &rootuser;, enter the &igw; container:
     </para>
--- a/xml/deploy_additional.xml
+++ b/xml/deploy_additional.xml
@@ -303,11 +303,7 @@
     <important>
      <title>Incorrectly reported image sizes and the <option>emulate_legacy_capacity</option> option</title>
      <para>
-      A bug in older kernels caused that an RBD target reported an image size
-      one block (512 bytes) larger than it actually was. Therefore, the LUN
-      could write one additional block of the image. To keep the image data
-      consistent, new kernels support the old behavior by default unless you
-      set <option>emulate_legacy_capacity</option> attribute to 0 for a target.
+      A bug in older kernels caused iSCSI targets to report LUNs as one block (512 bytes) larger than the underlying RBD image. This allowed iSCSI initiators to write to one additional block of the image. To keep the image data consistent, new kernels support the old behavior by default unless you set the <option>emulate_legacy_capacity</option> attribute to 0.
      </para>
      <para>
       We recommend disabling the old behavior for newly created targets. If you
--- a/xml/deploy_additional.xml
+++ b/xml/deploy_additional.xml
@@ -312,10 +312,10 @@
       <option>emulate_legacy_capacity</option> attribute to 0 for a target.
      </para>
      <para>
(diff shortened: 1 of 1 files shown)


xml/deploy_additional.xml
```diff
@@ -300,6 +300,55 @@
       </listitem>
      </itemizedlist>
     </note>
+    <important>
+     <title>Incorrectly reported image sizes</title>
```
trociny, line 304: I would have a title like "Disabling emulate_legacy_capacity option" or just "emulate_legacy_capacity option".

xml/deploy_additional.xml
```diff
@@ -300,6 +300,55 @@
...
+     <para>
+      A bug in older kernels caused that an RBD target reported an image size
+      one block (512B) larger than it actually was. Therefore, the LUN could
+      write one additional block of the image. Although this bug is fixed now,
+      existing images can be corrupted because they previously used the block
+      that is not accessible any more. For this reason, the new kernels support
```
trociny, line 310: Not sure about this sentence. It looks correct but might be confusing for users. @ddiss may be you have an idea how it could be changed?

trociny (approved):
In general LGTM. But I would like @ddiss look at this.

xml/deploy_additional.xml
```diff
@@ -300,6 +300,55 @@
       </listitem>
      </itemizedlist>
     </note>
+    <important>
+     <title>Incorrectly reported image sizes</title>
+     <para>
+      A bug in older kernels caused that an RBD target reported an image size
+      one block (512B) larger than it actually was. Therefore, the LUN could
+      write one additional block of the image. Although this bug is fixed now,
```
ddiss, line 308: nit: `Although this bug is now fixed, ...`

Also, I think `512B` should be spelled out as `512 bytes`.

xml/deploy_additional.xml
ddiss: Yeah, this seems a bit alarmist, although I'm not exactly sure how to best phrase it...

Corruption should never occur unless `emulate_legacy_capacity` is changed from the kernel default `emulate_legacy_capacity=1` to `emulate_legacy_capacity=0` without growing the image.

I.e. the safe options are:
- for newly created images that have never been exposed via iSCSI, disable `emulate_legacy_capacity` (`emulate_legacy_capacity=0`) during target creation (before it's exposed).
- for images that have been exposed via iSCSI with `emulate_legacy_capacity=1` (SES7+) or any version prior to SES7 (`emulate_legacy_capacity=1` behaviour was always enabled prior to the parameter appearing), there are two options:
  - do nothing; keep the `emulate_legacy_capacity=1` setting as-is
  - disconnect all initiators and take the iSCSI target offline. Grow the RBD image by 512 bytes, then explicitly set `emulate_legacy_capacity=0`. Finally, bring the iSCSI target back online.

The following would be dangerous:
- changing `emulate_legacy_capacity=1` to `emulate_legacy_capacity=0` without growing the image
    - initiators would lose access to the last block of the LUN, which could cause data corruption
- not disconnecting iSCSI initiators when performing the change
    - the `emulate_legacy_capacity=1->0` change and the RBD image 512 byte size increase needs to appear atomic from the SCSI initiator point of view

xml/deploy_additional.xml
```diff
@@ -300,6 +300,55 @@
...
+      &gwcli; shell to disable it:
+     </para>
+<screen>
+&prompt.root;&gwcli;
+&prompt.gwcli;cd /disks/<replaceable>POOL</replaceable>/<replaceable>IMAGE</replaceable>
+&prompt.gwcli;reconfigure emulate_legacy_capacity 0
+</screen>
+     <tip>
+      <para>
+       You can also disable the emulation in the &dashboard; when creating or
```
ddiss, line 327: I think `the parameter` would be a little clearer than `the emulation` here.

ddiss (changes requested):
Thanks very much for documenting this. I've suggested a few changes...

xml/deploy_additional.xml
ddiss: Perhaps just append it, e.g. `Incorrectly reported image sizes (emulate_legacy_capacity option)`?

tbazant, line 304:
```suggestion
     <title>Incorrectly reported image sizes and the <option>emulate_legacy_capacity</option> option</title>
```

xml/deploy_additional.xml
tbazant:
```suggestion
      write one additional block of the image. Although this bug is now fixed,
```

xml/deploy_additional.xml
```diff
@@ -300,6 +300,55 @@
       </listitem>
      </itemizedlist>
     </note>
+    <important>
+     <title>Incorrectly reported image sizes</title>
+     <para>
+      A bug in older kernels caused that an RBD target reported an image size
+      one block (512B) larger than it actually was. Therefore, the LUN could
```
tbazant, line 307:
```suggestion
      one block (512 bytes) larger than it actually was. Therefore, the LUN could
```

tbazant (commented):
I've rephrased a few sentences + made 2 suggestions

xml/deploy_additional.xml
```diff
@@ -300,6 +300,53 @@
...
      </itemizedlist>
     </note>
+    <important>
+     <title>Incorrectly reported image sizes and the <option>emulate_legacy_capacity</option> option</title>
+     <para>
+      A bug in older kernels caused that an RBD target reported an image size
+      one block (512 bytes) larger than it actually was. Therefore, the LUN
+      could write one additional block of the image. To keep the image data
+      consistent, new kernels support the old behavior by default unless you
+      set <option>emulate_legacy_capacity</option> attribute to 0 for a target.
```
ddiss, line 310: This looks good. I'd suggest a minor rework to avoid confusion between iSCSI targets, LUNs and images:

```
A bug in older kernels caused iSCSI targets to report LUNs as one block
(512 bytes) larger than the underlying RBD image. This allowed iSCSI
initiators to write to one additional block of the image. To keep the image
data consistent, new kernels support the old behavior by default unless you
set the <option>emulate_legacy_capacity</option> attribute to 0.
```

- emulate_legacy_capacity is an LIO backstore (LUN) attribute, i.e. it doesn't
  apply to the iSCSI target IQN, which could have multiple LUNs. I'm not sure
  how the UI presents it, but it'd make sense to use the same terms.

xml/deploy_additional.xml
```diff
@@ -300,6 +300,53 @@
...
+<screen>
+&prompt.root;&gwcli;
+&prompt.gwcli;cd /disks/<replaceable>POOL</replaceable>/<replaceable>IMAGE</replaceable>
+&prompt.gwcli;reconfigure emulate_legacy_capacity 0
+</screen>
+     <tip>
+      <para>
+       You can also disable the <option>emulate_legacy_capacity</option> option
+       in the &dashboard; when creating or editing the RBD target.
+      </para>
```
ddiss, line 327: It probably makes sense to state that the dashboard can be used for this before documenting the `gwcli` process, as the former should be a little more user friendly.

Also, I think it'd be good to have a warning here that the above steps should only be taken for new images which have *not* been exposed via iSCSI. Existing images require the rbd image to be resized...

xml/deploy_additional.xml
```diff
@@ -300,6 +300,53 @@
...
+      one block (512 bytes) larger than it actually was. Therefore, the LUN
+      could write one additional block of the image. To keep the image data
+      consistent, new kernels support the old behavior by default unless you
+      set <option>emulate_legacy_capacity</option> attribute to 0 for a target.
+     </para>
+     <para>
+      We recommend disabling the old behavior for newly created targets. If you
+      plan to enable the <option>object-map</option> feature for RBD images,
+      disabling <option>emulate_legacy_capacity</option> is required. Use the
+      &gwcli; shell to disable it:
```
ddiss, line 316: Also for LUN vs target clarity, I'd suggest:

```
We recommend disabling <option>emulate_legacy_capacity</option> for newly created RBD images when
setting up the iSCSI LUN. If you plan to enable the <option>object-map</option>
feature for RBD images, disabling <option>emulate_legacy_capacity</option> is
required.
```

xml/deploy_additional.xml
```diff
@@ -300,6 +300,53 @@
...
+      safely ignore related <literal>target_core_rbd: beyond EOD</literal>
+      kernel messages on corresponding gateway nodes.
+     </para>
+     <para>
+      If you want to disable <option>emulate_legacy_capacity</option> for
+      existing targets to avoid potential data corruption, we recommend
+      increasing the image size by on 512B to include the additional boundary
+      block. Resize the image when it is not used right before disabling the
+      emulation. To see the current image size in bytes, use the following
+      command:
```
ddiss, line 342: To make the disable-the-iscsi-target/lun procedure a little clearer, I'd suggest something like:

```
To avoid data corruption, <option>emulate_legacy_capacity</option> should
only be disabled on existing iSCSI LUNs if the RBD image is grown by one
512-byte block at the same time. Disconnect all iSCSI initiators and
disable the LUN or iSCSI target, grow the image using the procedure below,
disable <option>emulate_legacy_capacity</option> and then finally re-enable 
the LUN or iSCSI target.
```

* I'm not sure what the dashboard allows for WRT disabling individual LUNs
  or only an entire iSCSI target IQN, which may have a collection of LUNs.
  It'd make sense to change "disable the LUN or iSCSI target" to match that.

ddiss (changes requested):
Looks good. I have a few more minor suggestions. Thanks again

xml/deploy_additional.xml
```diff
@@ -300,6 +300,53 @@
...
+      <option>emulate_legacy_capacity</option> enabled. In such case you can
+      safely ignore related <literal>target_core_rbd: beyond EOD</literal>
+      kernel messages on corresponding gateway nodes.
+     </para>
+     <para>
+      If you want to disable <option>emulate_legacy_capacity</option> for
+      existing targets to avoid potential data corruption, we recommend
+      increasing the image size by on 512B to include the additional boundary
+      block. Resize the image when it is not used right before disabling the
+      emulation. To see the current image size in bytes, use the following
```
tbazant, line 341:
```suggestion
      To avoid data corruption, <option>emulate_legacy_capacity</option> should only be disabled on existing iSCSI LUNs if the RBD image is grown by one 512-byte block at the same time. Disconnect all iSCSI initiators and disable the LUN or iSCSI target, grow the image using the procedure below, disable <option>emulate_legacy_capacity</option> and then finally re-enable the LUN or iSCSI target. To see the current image size in bytes, use the following
```

tbazant (commented):
Created real GH suggestions

PR #1027, opened 2022-05-25
Adding new cluster services added (jsc#DOCTEAM-595)

### LINKED ISSUE #1010: Document deploying services via Dasboard
**Link To Section**
https://susedoc.github.io/doc-ses/main/html/ses-admin/dashboard-cluster.html#dashboard-cluster-services

**Summary**
Apart from viewing running services, you can now deploy them via Dashboard.

tbazant:
fixes #1010


lezare:
Although this in general looks good to me, two things:

1. When creating services via the dashboard we "expose" the `Labels` placement option, which in the documentation currently we only refer to specifying placement by host (using spec files, which is the recommended way for deploying services). We do not have any information in the documentation about using placement via labels (as far as I can tell).

This may be confusing for some customers when creating a service via the dashoard and seeing a `Placement` option `Label`.

So although not directly related here, should we consider adding information regarding placement using labels to the documentation? Maybe it can be as simple as adding a link to the upstream documentation [1]?

2. There also currently seems to be a problem that (at least on upgraded 6 > 7.1 clusters not using a local registry) the default registry is set to `docker.io` instead of the SUSE registry. It looks like also the dashboard then uses this registry when deploying services [2]. Not sure if we should wait with this PR until that is resolved.

[1] https://docs.ceph.com/en/quincy/cephadm/services/#placement-by-labels
[2] https://bugzilla.suse.com/show_bug.cgi?id=1200064

lezare (approved):
Registry problem is resolved with PR 480 from Bug 1200064 so we indeed can continue here.

LGTM.

xml/deploy_core.xml
```diff
@@ -235,6 +235,55 @@ placement:
   - host3
 [...]
 </screen>
+   <sect3 xml:id="cephadm-placement-specs-label">
+    <title>Placement by labels</title>
+    <para>
+     You can limit the placement of &ceph; services to hosts that match a
+     specific <emphasis>label</emphasis>. In the following example, we create a
+     label <literal>prometheus</literal>, assign it to specific hosts, and
+     apply the service &prometheus; to the group of hosts with
```
dariavladykina, line 244:
```suggestion
     apply the service &prometheus; to the group of hosts with the
```

dariavladykina (commented):
LGTM, just a minor nit, thanks

PR #1029, opened 2022-06-07
Re-run login to a custom registry (bsc#1196334)

tbazant:
fixes https://bugzilla.suse.com/show_bug.cgi?id=1196334


asettle (approved):

tserong (changes requested):
I've suggested one small wording tweak. Also, it's probably best to move this tip earlier in the steps in section 10.6, so it's _before_ any of the adopt commands, as those will all try to pull container images.

PR #1039, opened 2022-08-09
Added command to pull images from the default registry (bsc#122068)

tbazant:
fixes https://bugzilla.suse.com/show_bug.cgi?id=1202068


tserong (approved):
LGTM (note that we should be able to drop this again once we get our 16.2.9 update out)

PR #1040, opened 2022-08-10
Revert "Removed dangerous option during upgrade (bsc#1200883)"

asettle:
Setting bluestore_fsck_quick_fix_on_mount true in 16.2.7 has been okayed by @tserong

### DIFF
--- a/xml/admin_ceph_upgrade.xml
+++ b/xml/admin_ceph_upgrade.xml
@@ -516,7 +516,6 @@ Nodes running older software versions must be upgraded in the following order:
   </para>
 
   <procedure>
-   <!-- 2022-06-27 tbazant: not valid yet, see bsc#1200883
    <step>
     <para>
      Before adopting any OSD node, you need to perform a format conversion of
@@ -537,7 +536,6 @@ Nodes running older software versions must be upgraded in the following order:
      </para>
     </note>
    </step>
-   -->
    <step>
     <para>
      If the node you are upgrading is an OSD node, avoid having the OSD marked


tserong (approved):

tbazant:
@tserong How does this relate to https://github.com/SUSE/doc-ses/commit/e19a398f35d665ef1b3d1080f8f35b9248c3b299? If at all?

tserong:
> @tserong How does this relate to [e19a398](https://github.com/SUSE/doc-ses/commit/e19a398f35d665ef1b3d1080f8f35b9248c3b299)? If at all? See https://bugzilla.suse.com/show_bug.cgi?id=1201480 for more info.

Once this PR (to re-add `ceph config set osd bluestore_fsck_quick_fix_on_mount true`) goes in, then I think _most_ of e19a398 should be unnecessary as that's just a way of manually running a quick fix. The one thing I'm not sure about is the "compact" which is mentioned in that PR. Maybe @ifed01 can advise on that point?

ifed01:
> > @tserong How does this relate to [e19a398](https://github.com/SUSE/doc-ses/commit/e19a398f35d665ef1b3d1080f8f35b9248c3b299)? If at all? See https://bugzilla.suse.com/show_bug.cgi?id=1201480 for more info.
> 
> Once this PR (to re-add `ceph config set osd bluestore_fsck_quick_fix_on_mount true`) goes in, then I think _most_ of [e19a398](https://github.com/SUSE/doc-ses/commit/e19a398f35d665ef1b3d1080f8f35b9248c3b299) should be unnecessary as that's just a way of manually running a quick fix. The one thing I'm not sure about is the "compact" which is mentioned in that PR. Maybe @ifed01 can advise on that point?

+0.5 to @tserong.  
Once we have quick_fix_on_mount enabled this will convert OSD OMAPs on the next upgraded OSD's startup. Hence generally there no need to issue quick-fix manually afterwards. But on the other hand doing quick-fix manually is apparently a better approach as the upgrade script gets better control over the process, e.g. it explicitly controls when OMAP fixing is over.

As for compaction - indeed it worth running after(!) OMAP conversion/quick-fixing. The conversion could result in bulk data removal which is known for causing db access performance drop. Hence better to compact DB afterwards, offline compaction is preferred. 
Hence IMO the ideal upgrade procedure would look like (this omits steps to OMAP upgrade)
1) Shutdown legacy OSD
2) Upgrade Ceph binaries
3) Run ceph-bluestore-tool's quick-fix.
4) Run ceph-kvstore-tool's compact.
5) Start upgraded OSD (quick_fix_on_mount setting is irrelevant at this point, better use default IMO)

PR #1042, opened 2022-08-11
Update non-inclusive language in SES docs

tbazant:


asettle:
Note: Not everything from upstream has changed language - for this reason, not everything will be changed. This should be fine as is.

asettle (approved):

PR #1043, opened 2022-08-15
Added steps to migrate to new DB device (bsc#1202048)

tbazant:
fixes https://bugzilla.suse.com/show_bug.cgi?id=1202048

### DIFF
--- a/xml/admin_saltcluster.xml
+++ b/xml/admin_saltcluster.xml
@@ -1220,7 +1220,8 @@ OSD_ID  HOST         STATE                    PG_COUNT  REPLACE  FORCE  STARTED_
    <para>
     DB device belongs to an OSD and stores its metadata (see <xref linkend="about-bluestore"/> for
     more details). There are several reasons why you may want to migrate an existing DB device to a
-    new one - for example, when OSDs have different DB sizes and you need to align them.
+    new one&mdash;for example, when OSDs have different DB sizes and you need to 
+    align them.
    </para>
    <tip>
     <title><command>ceph-volume</command> naming convention</title>
--- a/xml/admin_saltcluster.xml
+++ b/xml/admin_saltcluster.xml
@@ -1215,6 +1215,86 @@ OSD_ID  HOST         STATE                    PG_COUNT  REPLACE  FORCE  STARTED_
     </para>
    </important>
   </sect2>
+  <sect2 xml:id="migrating-db-device">
+   <title>Migrating OSD's DB device</title>
+   <para>
+    DB device belongs to an OSD and stores its metadata (see <xref linkend="about-bluestore"/> for
+    more details). There are several reason why you may want to migrate an existing DB device to a
+    new one - for example, when OSDs have different DB sizes and you need to align them.
+   </para>
+   <procedure>
+    <title>Migrating a DB device to a new device</title>
+    <step>
+     <para>
+      Identify the <option>db device</option> and <option>osd fsid</option> values by running
+      the following command:
+     </para>
+<screen>&prompt.cephuser;cephadm ceph-volume lvm list
+[...]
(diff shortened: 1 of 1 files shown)


lezare (changes requested):
There is a small typo, `reason` should be `reasons` on line 1222, and the command sequence should be as per the following example:

1. `cephadm unit --name osd.${id} stop` on the relevant OSD node.
2. `cephadm shell --name osd.${id}` to enter into the stopped OSD container.
3. `ceph-volume lvm migrate --osd-id ${id} --osd-fsid ${fsid} --from db --target ceph-6ff5feb5-b489-4640-99c3-3551f881a307/osd-db-da349474-b33f-4efc-b4fc-a9c99a1d4e24`
4. `exit` the cephadm shell when done with the commands
5. `cephadm unit --name osd.${id} start` on the relevant OSD node to start the OSD again.

Should we maybe also add information on how to properly create a LV for the new DB? Something like the following command can be used to create the new LV with the desired size:

`lvcreate -n osd-db-$(cat /proc/sys/kernel/random/uuid) ceph-6ff5feb5-b489-4640-99c3-3551f881a307 --size <db_size>`

I am not sure if we should mention something (add a note?) about the ceph-volume naming convention, as this changed some over time (in case this is confusing for some customers). Some clusters may for example have VG / LV names as follows which differs from the current naming convention:

`ceph-block-dbs-<uuid>/osd-block-db-<uuid>` (old)
`ceph-<uuid>/osd-db-<uuid>` (current)

From a functional perspective, which ever naming convention is used for the VGs / LVs should in the end however not matter.

xml/admin_saltcluster.xml
```diff
@@ -1215,6 +1215,102 @@ OSD_ID  HOST         STATE                    PG_COUNT  REPLACE  FORCE  STARTED_
...
+     </para>
+<screen>&prompt.cephuser.osd;cephadm shell --name osd.0</screen>
+    </step>
+    <step>
+     <para>
+      Migrate the DB device:
+     </para>
+<screen>&prompt.cephuser;ceph-volume lvm migrate --osd-id 0 \
+ --osd-fsid c276d2a4-5578-4847-94c6-8e2e6abf81c4 --from db \
+ --target ceph-c3dc9227-ca3e-49bc-992c-00602cb3eec7/osd-db-b346b9ff-dbbe-40db-a95e-2419ccd31f2c</screen>
```
trociny, line 1293: @lezare  Don't we need `ceph-volume lvm new-db` before `lvm migrate`? I did not test these instructions but AFAIR when I did migration I needed to run `new-db`.

[1] https://docs.ceph.com/en/quincy/ceph-volume/lvm/newdb/

trociny: Or may be it was a little different case of migrating from osd withoud dedicated db...

Anyway I am fine with changes if they are tested and proved to work. Otherwise I will need to test it before approval.

xml/admin_saltcluster.xml
```diff
@@ -1215,6 +1215,102 @@ OSD_ID  HOST         STATE                    PG_COUNT  REPLACE  FORCE  STARTED_
     </para>
    </important>
   </sect2>
+  <sect2 xml:id="migrating-db-device">
+   <title>Migrating OSD's DB device</title>
+   <para>
+    DB device belongs to an OSD and stores its metadata (see <xref linkend="about-bluestore"/> for
+    more details). There are several reasons why you may want to migrate an existing DB device to a
+    new one - for example, when OSDs have different DB sizes and you need to align them.
```
dariavladykina, line 1223:
```suggestion
    new one&mdash;for example, when OSDs have different DB sizes and you need to 
    align them.
```

dariavladykina (commented):
LGTM!

PR #1044, opened 2022-09-08
Adjusting upgrade docs from 6 -> 7 (removing)

asettle:
We've had an issue (or two) with customers erroneously upgrading to version 7 when they can't. Deepsea won't allow the upgrade anymore and thus these docs are partially redundant. I'm open to suggestion if people have a better way to point our customers in the right direction.

Signed-off-by: Alexandra Settle <asettle@suse.com>


asettle:
It's complaining about two hyphens in the middle of the docs - I can simply remove them if we agree to the change, but otherwise, it's fine.

tbazant (approved):
Looks good (if it is true)

xml/admin_ceph_upgrade.xml
```diff
@@ -50,7 +53,7 @@
    supported. First, you must upgrade to the latest version of &productname;
    &prevproductnumber;, and then follow the steps in this chapter.
   </para>
- </important>
+<!---</important>
```
tbazant, line 56:
```suggestion
<!-- </important>
```

tbazant:
> Looks good (if it is true)

PR #1047, opened 2022-10-12
nfs: add setup instructions for HA NFS

m-ildefons:
Add setup instructions for HA NFS setup.

Signed-off-by: Moritz Röhrich <moritz.rohrich@suse.com>


m-ildefons:
Already included in PR#1049

PR #1050, opened 2022-10-19
Created new NFS HA and RGW HA graphics

janajaeger:
Check whether these convert OK. If not, report back.

Also, don't even try to include native SVG. Just go for PNG.


tbazant (approved):
Looks perfect

PR #1054, opened 2022-11-10
Switched order of steps when removing a node.

### LINKED ISSUE #1051: [doc] Issue in "Removing nodes"
### Type of report
   - [X] bug
   - [ ] feature request

### Source Document
https://documentation.suse.com/ses/7.1/single-html/ses-admin/#salt-node-removing

### Summary

The docs state to remove the remaining services (e.g. node-exporter, crash) after the host has been removed (step 2: "ceph orch host rm ses-min2"). This leads to an error:
```
storage01:~ # ceph orch host rm storage14
Error EINVAL: Not allowed to remove storage14 from cluster. The following daemons are running in the host:
type                 id             
-------------------- ---------------
node-exporter        storage14      
crash                storage14      

Please run 'ceph orch host drain storage14' to remove daemons from host

```
So it's required to remove all services first, then remove the host from ceph.

### Steps To Reproduce

Try removing a host according to the documentation:

1. Remove OSDs (draining)
2. Remove host from service specs
3. Remove host from ceph

### Expected Results

Host is removed.

### Actual Results

Host is not removed.

### Notes

Steps 2 and 3 should be switched.

tbazant:
fixes #1051

### DIFF
--- a/xml/admin_saltcluster.xml
+++ b/xml/admin_saltcluster.xml
@@ -167,12 +167,6 @@ placement:
    </step>
    <step>
     <para>
-     Remove the node from &cephadm;'s environment:
-    </para>
-<screen>&prompt.cephuser;ceph orch host rm ses-node2</screen>
-   </step>
-   <step>
-    <para>
      If the node is running <literal>crash.osd.1</literal> and
      <literal>crash.osd.2</literal> services, remove them by running the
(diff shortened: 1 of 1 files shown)


ebl-nde (approved):
looks good to me, I'll check next time if that works as expected.

PR #1057, opened 2022-11-28
Fixup `network.*` runner commands

mgfritch:
Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1204438

### DIFF
--- a/xml/bp_troubleshooting_tips.xml
+++ b/xml/bp_troubleshooting_tips.xml
@@ -434,9 +434,9 @@
    <title>Performing basic diagnostics</title>
    <para>
     Try the <command>salt-run network.ping</command> command to ping between
-    cluster nodes to see if individual interface can reach to specific
+    cluster nodes to see if an individual interface can reach to a specific
     interface and the average response time. Any specific response time much
-    slower then average will also be reported. For example:
+    slower than average will also be reported. For example:
    </para>
 <screen>&prompt.smaster;salt-run network.ping
 Succeeded: 8 addresses from 7 minions average rtt 0.15 ms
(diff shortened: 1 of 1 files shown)


xml/bp_troubleshooting_tips.xml
```diff
@@ -422,10 +422,18 @@
    verifying expectations.
   </para>
 
+  <tip>
+   <title>Sync Runners</title>
+   <para>
+    Use the <command>salt-run saltutil.sync_runners</command> command
+    if the salt runner is reported as <literal>not available</literal>.
```
tbazant, line 429:
```suggestion
    if the &salt; runner is reported as <literal>not available</literal>.
```

tbazant (approved):
one formal nit, otherwise OK

xml/bp_troubleshooting_tips.xml
```diff
@@ -422,10 +422,18 @@
...
+  </tip>
+
   <sect2 xml:id="network-test-basic-diagnostics">
    <title>Performing basic diagnostics</title>
    <para>
-    Try the <command>salt '*' network.ping</command> command to ping between
+    Try the <command>salt-run network.ping</command> command to ping between
     cluster nodes to see if individual interface can reach to specific
     interface and the average response time. Any specific response time much
     slower then average will also be reported. For example:
```
dariavladykina, line 439:
```suggestion
    cluster nodes to see if an individual interface can reach to a specific
    interface and the average response time. Any specific response time much
    slower than average will also be reported. For example:
```

dariavladykina (commented):
Hi, a small suggestion here. Thanks!

PR #1064, opened 2023-04-19
SES7: Removed duplicate step in a procedure

tbazant:


trociny (commented):
The current change looks OK, but actually this is a minor one. We need to remove the step that    says "Run the following commands to perform an offline compaction, enabling a quick fix, and  perform an online compaction on each OSD daemon that is deployed on the OSD node identified  by the <replaceable>MINION_ID</replaceable>."

We may just remove it or add as a note, that this manual fix could be done instead of setting `bluestore_fsck_quick_fix_on_mount` global config option (i.e. automatic fix).

@ifed01 please confirm

tbazant:
> The current change looks OK, but actually this is a minor one. We need to remove the step that says "Run the following commands to perform an offline compaction, enabling a quick fix, and perform an online compaction on each OSD daemon that is deployed on the OSD node identified by the MINION_ID."
> 
> We may just remove it or add as a note, that this manual fix could be done instead of setting `bluestore_fsck_quick_fix_on_mount` global config option (i.e. automatic fix).
@trociny made some changes, look if it's ok now

xml/admin_ceph_upgrade.xml
```diff
@@ -543,15 +543,38 @@ Nodes running older software versions must be upgraded in the following order:
...
-            <link
-      xlink:href="https://docs.ceph.com/en/latest/releases/pacific/#upgrading-non-cephadm-clusters"/>.
+            Alternatively, you can set the
+            <option>bluestore_fsck_quick_fix_on_mount</option> option for each
+            OSD daemon that is deployed on the rebooted node.
+          </para>
+<screen>&prompt.smaster;ceph config set osd.<replaceable>OSD_DAEMON_ID</replaceable> \
+ bluestore_fsck_quick_fix_on_mount true</screen>
+          <para>
+            In such case, perform an offline compaction enabling a quick fix
```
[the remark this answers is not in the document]
tbazant: @trociny it's getting a bit complicated, how about just removing the alternative per-OSD way and leave the global one only?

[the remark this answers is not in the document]
trociny: Agree, I don't think it is worth much mentioning.

trociny (approved):
LGTM

xml/admin_ceph_upgrade.xml
```diff
@@ -543,15 +543,31 @@ Nodes running older software versions must be upgraded in the following order:
...
           <para>
-            The conversion may take minutes to hours, depending on how much
-            OMAP data the related hard disk contains. For more details, refer
-            to
-            <link
-      xlink:href="https://docs.ceph.com/en/latest/releases/pacific/#upgrading-non-cephadm-clusters"/>.
+            You can also perform an offline compaction enabling a quick fix
+            and perform an online compaction on each OSD daemon that is
+            deployed on the OSD node identified by the
+            <replaceable>MINION_ID</replaceable>. For example
```
dariavladykina, line 556:
```suggestion
            <replaceable>MINION_ID</replaceable>. For example,
```

xml/admin_ceph_upgrade.xml
```diff
@@ -605,51 +621,11 @@ ID   CLASS  WEIGHT   TYPE NAME       STATUS  REWEIGHT  PRI-AFF
       </step>
       <step>
         <para>
-          If the node you are upgrading is an OSD node, follow these steps:
+          If the node you are upgrading is an OSD node, then after the OSD node
+          with the &salt; minion ID <replaceable>MINION_ID</replaceable> has
+          been rebooted and now is up, run the following command:
```
dariavladykina, line 626:
```suggestion
          If the node you are upgrading is an OSD node, then, after the OSD node
          with the &salt; minion ID <replaceable>MINION_ID</replaceable> has
          been rebooted and is now up, run the following command:
```

dariavladykina (commented):
LG, just minor nits. Thanks!

xml/admin_ceph_upgrade.xml
```diff
@@ -543,15 +543,31 @@ Nodes running older software versions must be upgraded in the following order:
...
-            <link
-      xlink:href="https://docs.ceph.com/en/latest/releases/pacific/#upgrading-non-cephadm-clusters"/>.
+            You can also perform an offline compaction enabling a quick fix
+            and perform an online compaction on each OSD daemon that is
+            deployed on the OSD node identified by the
+            <replaceable>MINION_ID</replaceable>. For example
+          </para>
+<screen>
+&prompt.cephuser;cephadm unit --name osd.<replaceable>OSD_DAEMON_ID</replaceable> stop
+&prompt.cephuser;ceph-kvstore-tool bluestore-kv /var/lib/ceph/263302b4-afa2-11ec-a986-525400af093c/osd.<replaceable>OSD_DAEMON_ID</replaceable> compact
```
tserong, line 560: This would be better as:

```suggestion
&prompt.cephuser;cephadm shell --name osd.<replaceable>OSD_DAEMON_ID</replaceable> ceph-kvstore-tool bluestore-kv /var/lib/ceph/osd/ceph-<replaceable>OSD_DAEMON_ID</replaceable> compact
```

(just because it means you don't have to _also_ replace that irritating FSID with whatever the actual cluster FSID is)

xml/admin_ceph_upgrade.xml
```diff
@@ -543,15 +543,31 @@ Nodes running older software versions must be upgraded in the following order:
...
+&prompt.cephuser;cephadm shell --name osd.<replaceable>OSD_DAEMON_ID</replaceable> ceph-bluestore-tool --path /var/lib/ceph/osd/ceph-<replaceable>OSD_DAEMON_ID</replaceable>--command quick-fix
```
tserong, line 561: Missing space before `--command`:

```suggestion
&prompt.cephuser;cephadm shell --name osd.<replaceable>OSD_DAEMON_ID</replaceable> ceph-bluestore-tool --path /var/lib/ceph/osd/ceph-<replaceable>OSD_DAEMON_ID</replaceable> --command quick-fix
```

xml/admin_ceph_upgrade.xml
```diff
@@ -543,15 +543,31 @@ Nodes running older software versions must be upgraded in the following order:
...
+&prompt.cephuser;cephadm unit --name osd.<replaceable>OSD_DAEMON_ID</replaceable> start
+</screen>
+          <para>
+            Allow the OSD node to be up and in again, then run:
+          </para>
+<screen>&prompt.cephuser;cephadm enter --name osd.<replaceable>OSD_DAEMON_ID</replaceable> ceph daemon osd.<replaceable>OSD_DAEMON_ID</replaceable> compact</screen>
```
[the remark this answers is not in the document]
tserong: I'm honestly not sure - I think the original commands came from Boyd in https://bugzilla.suse.com/show_bug.cgi?id=1201480, but that second compact does look redundant to me.

I also noticed an old comment from Igor https://github.com/SUSE/doc-ses/pull/1040#issuecomment-1243618845 which says to do the quick-fix _before_ the compact, so the previous steps (`ceph-kvstore-tool bluestore-kv [...]` and `ceph-bluestore-tool [...] --command quick-fix`) should probably be swapped, but I'd be happiest if @ifed01 could also review this.

xml/admin_ceph_upgrade.xml
```diff
@@ -543,15 +543,31 @@ Nodes running older software versions must be upgraded in the following order:
 <screen>&prompt.cephuser;ceph config set osd bluestore_fsck_quick_fix_on_mount true</screen>
         <para>
           The OSD nodes will be converted automatically after their adoption
-          finishes.
+          finishes. The conversion may take minutes to hours, depending on how
+          much OMAP data the related hard disk contains. For more details, refer
```
ifed01, line 547: May be replace 'related hard disk' with 'related disk'?

xml/admin_ceph_upgrade.xml
```diff
@@ -543,15 +543,31 @@ Nodes running older software versions must be upgraded in the following order:
 <screen>&prompt.cephuser;ceph config set osd bluestore_fsck_quick_fix_on_mount true</screen>
```
ifed01, line 543: The above text notes that this command performs format conversion.  IMO this is not 100% correct. The command just enables that format conversion on the first start of an adopted OSD.

xml/admin_ceph_upgrade.xml
```diff
@@ -543,15 +543,31 @@ Nodes running older software versions must be upgraded in the following order:
...
+          <link xlink:href="https://docs.ceph.com/en/latest/releases/pacific/#upgrading-non-cephadm-clusters"/>.
         </para>
         <note>
           <para>
-            The conversion may take minutes to hours, depending on how much
-            OMAP data the related hard disk contains. For more details, refer
-            to
-            <link
-      xlink:href="https://docs.ceph.com/en/latest/releases/pacific/#upgrading-non-cephadm-clusters"/>.
+            You can also perform an offline compaction enabling a quick fix
```
ifed01, line 553: Generally this is a redundant section. In the current form it rather describes an alternative way of doing omap conversion manually if the previous step (setting bluestore_fsck_quick_fix_on_mount to true)  is omitted for whatever reason.  Which we apparently do not expect.
On the other hand we might want to enforce users to do omap conversion manually - in this case the above step (setting bluestore_fsck_quick_fix_on_mount to true) which enables automatic omap conversion is redundant.
I'm leaning towards recommending manual conversion approach as it allows finer control over the process. So if there are no objections the resulting text should look like:
"
 Before adopting any OSD node, you need to perform a format conversion
 of OSD nodes to improve the accounting for OMAP data. You can do so
 by running the following commands on the &adm;:
&prompt.cephuser;cephadm unit --name osd.<replaceable>OSD_DAEMON_ID</replaceable> stop
&prompt.cephuser;cephadm shell --name osd.<replaceable>OSD_DAEMON_ID</replaceable> ceph-bluestore-tool --path /var/lib/ceph/osd/ceph-<replaceable>OSD_DAEMON_ID</replaceable> --command quick-fix
&prompt.cephuser;cephadm unit --name osd.<replaceable>OSD_DAEMON_ID</replaceable> start

The conversion may take minutes to hours, depending on how
much OMAP data the related hard disk contains. For more details, refer <...>
 You can run the above commands in parallel on multiple OSD daemons
 on the same OSD node to help accelerate the upgrade.
"

xml/admin_ceph_upgrade.xml
```diff
@@ -543,15 +543,31 @@ Nodes running older software versions must be upgraded in the following order:
...
+            and perform an online compaction on each OSD daemon that is
+            deployed on the OSD node identified by the
+            <replaceable>MINION_ID</replaceable>. For example,
+          </para>
+<screen>
+&prompt.cephuser;cephadm unit --name osd.<replaceable>OSD_DAEMON_ID</replaceable> stop
+&prompt.cephuser;cephadm shell --name osd.<replaceable>OSD_DAEMON_ID</replaceable> ceph-kvstore-tool bluestore-kv /var/lib/ceph/osd/ceph-<replaceable>OSD_DAEMON_ID</replaceable> compact
```
ifed01, line 560: This is redundant

xml/admin_ceph_upgrade.xml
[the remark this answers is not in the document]
ifed01: Yes, this is a comletely redundant action since quick-fix does compaction itself

PR #1071, opened 2024-03-07
Add metadata for SES in book  & article files

tomschr:
* Change version attribute to "5.2"
* Add `<meta name="series">`
* Add `<meta name="task">` (sometimes disabled)
* Add the ITS namespace in root element


tbazant (approved):
Looks good, thanks Toms

